PR #1, opened 2017-09-25
working on api spec

pupapaik:


api.yml
```diff
@@ -0,0 +1,203 @@
...
+    type: object
+    properties:
+      id:
+        type: "integer"
+        format: "int64"
+      name:
+        type: "string"
+      provisioner:
+        type: "integer"
+        format: "int64"
```
tomkukral: Is provisioner really int64?

api.yml
```diff
@@ -0,0 +1,203 @@
...
+          - "deployed"
+          - "pending"
+          - "failed"
+          - "deleted"
+  ClusterNode:
+    type: object
+    properties:
+      name:
+        type: string
+      # TODO: more properties
```
tomkukral: I'd suggest to dont' add more properties, lets use Kuberneties native node.

api.yml
```diff
@@ -0,0 +1,203 @@
...
+  ClusterService:
+    type: object
+    properties:
+      name:
```
tomkukral: Once more, lets use native Kubernetes services.

api.yml
```diff
@@ -0,0 +1,203 @@
...
+        type: string
+      # TODO: more properties
+  Provisioner:
+    type: object
+    properties:
+      id:
+        type: "integer"
```
tomkukral: What does this id means?

api.yml
```diff
@@ -0,0 +1,203 @@
...
+  - http
+
+basePath: /api/v1
+
+definitions:
+  Cluster:
+    type: object
+    properties:
+      id:
+        type: "integer"
```
Frostman: usually it's better not to use int as id. some uuid (string) or human-readable (string as well) will be more flexible.

api.yml
fpytloun: It's id of linked provisioner object

api.yml
fpytloun: unique identifier of that object, or we can use names as unique identifiers and don't use ids at all

Frostman: ID and name aren't the same - if you delete object with some name and than create new one with same name, the ID should be changed, so, you can differentiate the objects.

api.yml
tomkukral: uuid is a good idea, thanks!

PR #3, opened 2017-09-26
base provisioner

tomkukral:
Create class Provisioner as a base class for all provisioners.


kqueen/blueprints/ui/views.py
```diff
@@ -107,18 +108,22 @@ def provisioner_create():
             provisioner = Provisioner(
                 name=form.name.data,
                 engine=form.engine.data,
-                state='Not Available',
-                location='-',
-                access_id=form.access_id.data,
-                access_key=form.access_key
+                state=app.config['PROVISIONER_UNKNOWN_STATE'],
+                parameters=json.loads({
```
tomkukral: `json.loads` should be here, just `dict`.

kqueen/blueprints/ui/views.py
```diff
@@ -107,18 +108,22 @@ def provisioner_create():
...
+                    'username': form.username.data,
+                    'password': form.password.data
+                })
             )
             # Check if provisioner lives
-            if provisioner.alive():
-                provisioner.state.value = 'OK'
+            if provisioner.alive() == True:
```
tomkukral: This should be in the model IMHO

kqueen/models.py
```diff
@@ -119,8 +101,11 @@ def engine_cls(self):
 
     @property
     def engine_name(self):
-        return self.engine_cls.__name__ if self.engine_cls else self.engine.value
+        return self.get_engine_cls().__name__ if self.get_engine_cls() else self.engine.value
```
tomkukral: Readability first :smile:

tomkukral:
What about some tests? @atengler

PR #26, opened 2017-10-11
Count used resources

tomkukral:
Goal of this change it provide function to sum all requested (and limited) resources per node.

This information will be used in UI to give users better visualization of cluster load.


kqueen/kubeapi.py
```diff
@@ -92,6 +93,49 @@ def count_pods_by_node(self):
...
+                    if resources:
+                        for resource_policy in ['limits', 'requests']:
+                            policy = resources.get(resource_policy, {})
+
+                            if policy:
+                                for resource_type in ['cpu', 'memory']:
+                                    value = policy.get(resource_type)
+
+                                    if value:
+                                        out[node_name][resource_policy][resource_type] += prefix_to_num(value)
```
tomkukral: Any idea how to write this crazy `for` jungle in better way?

tomkukral: Any idea how to write this crazy for jungle in better way?

PR #32, opened 2017-10-12
Add particles.json

atengler:


kqueen/asset/dynamic/js/particles.json
```diff
@@ -0,0 +1,110 @@
+{
```
tomkukral: Can we move ui related assets to ui blueprint?

PR #119, opened 2017-11-09
Swagger 2.0 and 3.0 API definitions and Swagger UI

atengler:


requirements.txt
```diff
@@ -1,5 +1,6 @@
 Flask==0.12.2
 Flask-JWT==0.3.2
+git+https://github.com/atengler/flask-swagger-ui.git@3.4.3-update#flask_swagger_ui
```
tomkukral: Fuj, is this really necessary? We need to have this in `setup.py` as well.

kqueen/blueprints/api/api2_0.yml
```diff
@@ -0,0 +1,791 @@
+swagger: "2.0"
```
tomkukral: Do we really need to have both versions? I know there are some advantages of v 3.0 but is it worth maintaining two versions?

atengler: 3.0 describes the API better, because of the improved format and we should use it after all swagger-api subprojects supports it, but as for now its supported by swagger-ui, but not by swagger-codegen. So we have 3.0 for the future use and current use with swagger-ui and 2.0 for swagger-codegen to potentially generate API clients for the supported languages.

requirements.txt
atengler: We can get rid of this after this is merged, as we discussed yesterday - https://github.com/sveint/flask-swagger-ui/pull/14

PR #120, opened 2017-11-10
Configurable host address and port

samos123:
Make the listen host address and port configurable


tomkukral:
Thank you @samos123 for your PR!

We are using `gunicorn` to run kqueen in production/contain and it ignores this options. However, it is certainly needed for development. I'll merge it right after #119

PR #126, opened 2017-11-13
ETCD storage DatetimeField

atengler:
This PR adds DatetimeField class to ETCD storage module. DatetimeField stores datetimes as UTC timestamps (floating point numbers) on the backend and returns them as naïve datetime.

### DIFF
--- a/kqueen/config/base.py
+++ b/kqueen/config/base.py
@@ -8,10 +8,6 @@ class BaseConfig:
     KQUEEN_HOST = '127.0.0.1'
     KQUEEN_PORT = 5000
 
-    # Babel default settings
-    BABEL_DEFAULT_TIMEZONE = 'UTC'
-    BABEL_DEFAULT_LOCALE = 'en'
-
     # etcd settings
     # ETCD_HOST
     # ETCD_PORT
--- a/kqueen/config/base.py
+++ b/kqueen/config/base.py
@@ -8,6 +8,10 @@ class BaseConfig:
     KQUEEN_HOST = '127.0.0.1'
     KQUEEN_PORT = 5000
 
+    # Babel default settings
+    BABEL_DEFAULT_TIMEZONE = 'UTC'
+    BABEL_DEFAULT_LOCALE = 'en'
+
     # etcd settings
     # ETCD_HOST
     # ETCD_PORT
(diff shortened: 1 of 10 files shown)


kqueen/config/base.py
```diff
@@ -8,6 +8,10 @@ class BaseConfig:
     KQUEEN_HOST = '127.0.0.1'
     KQUEEN_PORT = 5000
 
+    # Babel default settings
+    BABEL_DEFAULT_TIMEZONE = 'UTC'
```
tomkukral: We should probably keep backend on UTC everytime.

kqueen/models.py
```diff
@@ -30,6 +31,8 @@ class Cluster(Model, metaclass=ModelMeta):
     state = StringField()
     kubeconfig = JSONField()
     metadata = JSONField()
+    created_at = DatetimeField()
+    deleted_at = DatetimeField()
```
tomkukral: Do you plan set `deleted_at` instead of deleting objects? It won't make sense IMO.

kqueen/server.py
```diff
@@ -1,4 +1,5 @@
 from flask import Flask
+from flask.ext.babel import Babel
```
tomkukral: Can we do same with `python-dateutil` and `pytz`?

requirements.txt
```diff
@@ -1,4 +1,5 @@
 Flask==0.12.2
+Flask-Babel=0.11.2
```
tomkukral: Should be `Flask-Babel==0.11.2`

setup.py
```diff
@@ -21,6 +21,7 @@
     zip_safe=False,
     install_requires=[
         'Flask==0.12.2',
+        'Flask-Babel=0.11.2',
```
tomkukral: Should be `Flask-Babel==0.11.2`

kqueen/models.py
atengler: I will remove deleted_at field

kqueen/server.py
atengler: Not with the same functionality. But if you want to keep backend on UTC always, there has to be date parsing and transformation in API client in the UI, so I might as well only return the UTC timestamp and move the smart stuff to UI.

PR #132, opened 2017-11-16
User Active

atengler:
Add BoolField to ETCD storage module and add active attribute to User model.

### DIFF
--- a/devenv.py
+++ b/devenv.py
@@ -43,7 +43,8 @@ with app.app_context():
             username='admin',
             password='default',
             organization=organization,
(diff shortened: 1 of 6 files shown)


tomkukral:
What do you think about using more general field `user_status` . I may cover more states, for example:

* 0 - not_active
* 1 - active
* 2 - suspended
* 3 - admin

atengler:
Feels like mixing two things together, I would introduce role attribute for User model instead and keep active attribute for it is supposed to be.

PR #135, opened 2017-11-21
Engine Parameters

atengler:
Improve handling of engine parameters and expose them in API

### DIFF
--- a/kqueen/blueprints/api/views.py
+++ b/kqueen/blueprints/api/views.py
@@ -1,10 +1,12 @@
 from .helpers import get_object
 from flask import abort
 from flask import Blueprint
+from flask import current_app as app
 from flask import jsonify
 from flask import make_response
 from flask_jwt import current_identity
 from flask_jwt import jwt_required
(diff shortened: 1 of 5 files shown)


kqueen/blueprints/api/views.py
```diff
@@ -217,6 +219,31 @@ def user_whoami():
     return jsonify(output)
 
 
+@api.route('/engines', methods=['GET'])
+def engine_list():
+    engines = app.config.get('KQUEEN_ENGINES', [])
```
tomkukral: what about `from kqueen.engines.__init__ import __all__ as engines`?

kqueen/blueprints/api/views.py
```diff
@@ -217,6 +219,31 @@ def user_whoami():
     return jsonify(output)
 
 
+@api.route('/engines', methods=['GET'])
```
tomkukral: I was thinking about `/provisioners/engines` but `/engines` works for me too.

PR #140, opened 2017-11-23
Azure Managed Kubernetes Cluster

pupapaik:
AKS provisioner for managed kubernetes cluster.

### DIFF
--- a/kqueen/engines/test_aks.py
+++ b/kqueen/engines/test_aks.py
@@ -2,12 +2,10 @@ from .aks import AksEngine
 
 
 class TestAksInit:
-    def setup(self):
-        pass
 
     def test_init_aks(self, cluster, monkeypatch):
         def fake_client(self):
             return True
 
-        monkeypatch.setattr(AksEngine, "_get_client", fake_client)
+        monkeypatch.setattr(AksEngine, '_get_client', fake_client)
         AksEngine(cluster)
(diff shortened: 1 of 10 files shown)


kqueen/engines/test_aks.py
```diff
@@ -0,0 +1,13 @@
+from .aks import AksEngine
+
+
+class TestAksInit:
+    def setup(self):
```
tomkukral: This setup can be deleted.

kqueen/engines/test_aks.py
```diff
@@ -0,0 +1,13 @@
...
+        pass
+
+    def test_init_aks(self, cluster, monkeypatch):
+        def fake_client(self):
+            return True
+
+        monkeypatch.setattr(AksEngine, "_get_client", fake_client)
```
tomkukral: This isn't a blockt but we are trying to uses consistent quotes - `'`

run_aks_test.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python3
```
tomkukral: Is this file required to be merged into master?

setup.py
```diff
@@ -44,6 +44,8 @@
         'google-api-python-client==1.6.4',
         'google-auth==1.2.1',
         'google-auth-httplib2==0.0.3',
+        'azure',
```
tomkukral: Maybe we should pin versions?

run_aks_test.py
pupapaik: it is example for aks testing

PR #148, opened 2017-12-01
Config changes

atengler:
Change configuration to fix demo docker compose and introduce the possibility to bootstrap admin user during container startup.


tomkukral:
What about adding some documentation how to use it?

PR #149, opened 2017-12-01
Change handling of cluster state

atengler:
Our solution is still synchronous at the moment, so we should at least update the state every time Cluster resource is requested.

### DIFF
--- a/kqueen/blueprints/api/views.py
+++ b/kqueen/blueprints/api/views.py
@@ -62,6 +62,13 @@ def index():
 class ListClusters(ListView):
     object_class = Cluster
 
+    def get_content(self, *args, **kwargs):
+        clusters = super(ListClusters, self).get_content(*args, **kwargs)
+        # refresh states on all clusters
+        # TODO: replace this with some async function
+        [c.get_state() for c in clusters]
+        return clusters
+
 
 class CreateCluster(CreateView):
     object_class = Cluster
--- a/kqueen/blueprints/api/views.py
+++ b/kqueen/blueprints/api/views.py
@@ -15,12 +15,14 @@ from kqueen.models import Cluster
 from kqueen.models import Organization
 from kqueen.models import Provisioner
 from kqueen.models import User
+from kqueen.config import current_config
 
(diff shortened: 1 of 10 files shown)


kqueen/blueprints/api/views.py
```diff
@@ -62,6 +62,13 @@ def index():
 class ListClusters(ListView):
     object_class = Cluster
 
+    def get_content(self, *args, **kwargs):
```
tomkukral: I'll add tests for this.

kqueen/blueprints/api/views.py
```diff
@@ -62,6 +62,13 @@ def index():
...
+        clusters = super(ListClusters, self).get_content(*args, **kwargs)
+        # refresh states on all clusters
+        # TODO: replace this with some async function
+        [c.get_state() for c in clusters]
```
tomkukral: Do we really want to run this on all clusters? I'm afraid this call can be really slow.

atengler: It slows down the overview page loading, but its the only way to get the page to update the states correctly at the momment

tomkukral: Understand ... I'l try to use some kind of background threading for this call since we probably don't need synchronous update.

tomkukral: It isn't possible to use threading because the we'll get out of appication context. I'll try to do more hacking about it but we may stick with your solution as it may be the only way to go.

kqueen/blueprints/api/views.py
```diff
@@ -62,6 +67,27 @@ def index():
 class ListClusters(ListView):
     object_class = Cluster
 
+    async def _update_clusters(self, clusters):
+        loop = asyncio.get_event_loop()
+        futures = [loop.run_in_executor(None, c.get_state) for c in clusters]
+
+        for state in await asyncio.gather(*futures):
+            print(state)
```
tomkukral: Do somethins more useful, maybe logging or just pass.

PR #150, opened 2017-12-01
Docker compose update

atengler:
Add kqueen_mail to demo compose and fixate version of UI


docker-compose.demo.yml
```diff
@@ -14,8 +14,15 @@ services:
...
+  kqueen_mail:
+    image: modularitycontainers/postfix
+    volumes:
+      - /var/spool/postfix:/var/spool/postfix
+      - /var/spool/mail:/var/spool/mail
+    environment:
+      MYHOSTNAME: 'kqueen_mail'
   kqueen_ui:
-    image: atengler/kqueen-ui
+    image: atengler/kqueen-ui:v0.1
```
tomkukral: Can we use `kqueen/ui:v0.1` it should be same image. It will look better.

PR #159, opened 2017-12-06
WIP: Jenkins job parameters

tomkukral:
Don't delete stack after deploy.


kqueen/config/prod.py
```diff
@@ -13,7 +13,7 @@ class Config(BaseConfig):
     # Jenkins engine settings
     JENKINS_API_URL = 'https://ci.mcp.mirantis.net'
     JENKINS_PROVISION_JOB_NAME = 'deploy-aws-k8s_ha_calico_sm'
-    JENKINS_PROVISION_JOB_CTX = {}
+    JENKINS_PROVISION_JOB_CTX = {'STACK_DELETE': False}
```
atengler: Please, don't.

tomkukral: Why? I doesn't work without this parameter bacause cluster will be deleted right after deployment. We have already hardcoded Jenkins URL and job name.

PR #160, opened 2017-12-06
KQueen RBAC

atengler:
This PR implements basic solution for RBAC for KQueen app. `DEFAULT_POLICIES` app.config key is set by reading `default_policies.json` file, its possible to override these keys by setting `policy` attribute on Organization object. Relevant policy is picked in method on `kqueen.blueprints.api.generic_views.GenericView` and evaluated by `kqueen.auth.is_authorized` helper method.


kqueen/config/default_policy.json
```diff
@@ -0,0 +1,22 @@
+{
+    "cluster:create": ["admin"],
+    "cluster:delete": ["admin"],
+    "cluster:get": "all",
+    "cluster:list": "all",
+    "cluster:update": "all",
+    "organization:create": ["superadmin"],
+    "organization:delete": ["superadmin"],
+    "organization:get": "all",
+    "organization:list": "all",
```
tomkukral: Maybe `organizatoin:list` just superadmin ... do we need it somewhere else? Is this capability required for duplicates in namespaces?

kqueen/auth.py
```diff
@@ -40,3 +42,40 @@ def identity(payload):
...
+    Args:
+        user (dict or User): User data
+        policy_value (string or list): Either "all" or list of allowed roles
+
+    Returns:
+        bool: authorized or not
+    """
+    if isinstance(user, User):
+        role = user.get_dict()['role']
+    elif isinstance(user, dict):
```
tomkukral: Why do we need to support passing user as `dict`. I'd prefer to pass `User` object everywhere.

atengler: Because what we have at hand in API is not actually User object. I tried it and for whatever reason current_identity is not actually instance of User (even though it definitely looks that way), its some kind of werkzeug proxy instance. So I have to pass current_identity.get_dict() as argument for this method in generic_views.

kqueen/config/default_policy.json
atengler: I think it is, but when I do out of scope API calls from UI, I use service account, which has superadmin privileges anyways, so it should be possible to restrict organization:list to superadmin only. I'll try it tomorrow.

kqueen/auth.py
tomkukral: Thanks, got it.

kqueen/blueprints/api/generic_views.py
```diff
@@ -17,8 +18,40 @@ def get_class(self):
...
+        policy_value = policies.get(policy, None)
+
+        # evaluate user permissions
+        if policy_value and current_identity:
+            user = current_identity.get_dict()
+            authorized = is_authorized(user, policy_value)
+            if not authorized:
+                raise JWTError('Insufficient permissions',
+                               'Your user account is lacking the necessary '
+                               'permissions to perform this operation')
```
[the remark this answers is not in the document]
atengler: Not actually outdated comment, mb. OK, I'll add logger.warning tomorrow.

atengler:
I separated the code which picks relevant policy rule and code which actually evaluates if given action is permitted for given user. So we can easily create new rules and implement them in `is_authorized` method. So I can implement some new rules tomorrow, shouldn't be too much trouble.

PR #165, opened 2017-12-12
Setup new event loop and pass it to async method

atengler:
Current implementation of asyncio loop doesn't seem to be working for me, I tried to fix it this way.


tomkukral:
It is working when running in gunicorn, please let me check this.

tomkukral:
@atengler Can you share link to some more information about it?

PR #166, opened 2017-12-12
PasswordField

atengler:
Implement on_create method for Field base class and PasswordField (one-way hashed field).

### DIFF
--- a/kqueen/auth.py
+++ b/kqueen/auth.py
@@ -3,6 +3,7 @@
 from kqueen.models import Organization, User
 from uuid import uuid4
 
+import bcrypt
 import logging
 
 logger = logging.getLogger(__name__)
@@ -20,11 +21,12 @@ def authenticate(username, password):
         user: authenticated user
 
     """
-    from kqueen.server import bcrypt
     users = list(User.list(None, return_objects=True).values())
     username_table = {u.username: u for u in users}
     user = username_table.get(username)
-    if user and user.active and bcrypt.check_password_hash(user.password, password):
+    user_password = user.password.encode('utf-8')
+    given_password = password.encode('utf-8')
+    if user and user.active and bcrypt.checkpw(given_password, user_password):
         return user
 
 
--- a/kqueen/auth.py
+++ b/kqueen/auth.py
@@ -21,10 +21,11 @@ def authenticate(username, password):
         user: authenticated user
 
     """
+    from kqueen.server import bcrypt
     users = list(User.list(None, return_objects=True).values())
     username_table = {u.username: u for u in users}
     user = username_table.get(username)
-    if user and user.active and safe_str_cmp(user.password.encode('utf-8'), password.encode('utf-8')):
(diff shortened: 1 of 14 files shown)


kqueen/auth.py
```diff
@@ -16,10 +15,11 @@ def authenticate(username, password):
         user: authenticated user
 
     """
+    from kqueen.server import bcrypt
```
tomkukral: Is there a reason to import it here and not at the the beginning?

kqueen/blueprints/api/test_user.py
```diff
@@ -34,7 +34,7 @@ def get_auth_headers(self):
         Returns:
             dict: Authorization header dict
         """
-        data = {'username': self.obj.username, 'password': self.obj.password}
+        data = {'username': self.obj.username, 'password': self.obj.username + 'password'}
```
tomkukral: Why do we need to pass self.obj.username twice?

kqueen/auth.py
atengler: I think there was some circular import problem, import failed at the beginning of file.

kqueen/blueprints/api/test_user.py
atengler: Well, we don't. I removed the fake password, because it's random, there is no simple way to pass it along with the user object and `user.password` no longer returns value you could use for getting auth header. So I changed test users password to his username + the string 'password'. We can find another way, but this was quick fix for getting auth header in tests.

tomkukral: Oh, it's just test. Fine, I understand it now.

kqueen/conftest.py
```diff
@@ -23,6 +23,7 @@
 def app():
     """Prepare app."""
     app = create_app()
+    app.testing = True
```
tomkukral: Can you comment what is this doing?

kqueen/server.py
```diff
@@ -55,6 +56,8 @@ def create_app(config_file=None):
 
 
 app = create_app()
+# setup Bcrypt
+bcrypt = Bcrypt(app)
```
tomkukral: This is probably causing circual depency mentioned above. This initialization of `Bcrypt` in server module looks insane to me.

What is advantage of using `flask-bcrypt` instead of bare `bcrypt`? 

https://pypi.python.org/pypi/bcrypt/3.1.0 - sectip password hashing. This approach don't need this silly "global" variables.

kqueen/conftest.py
atengler: According to Flask docs it should be set to True in test app, possibly so you can test inside the actual project code if it was ran by test case. Not sure if we actually need it, but should be set to True I guess?

kqueen/server.py
```diff
@@ -55,6 +55,7 @@ def create_app(config_file=None):
 
 
 app = create_app()
+# setup Bcrypt
```
tomkukral: Remove this.

PR #167, opened 2017-12-13
Cluster provisioning state API endpoint

atengler:
Endpoint for getting status of cluster provisioning, returns result of `get_progress` method on relevant Engine object.

### DIFF
--- a/kqueen/blueprints/api/views.py
+++ b/kqueen/blueprints/api/views.py
@@ -156,6 +156,27 @@ def cluster_kubeconfig(pk):
     return jsonify(obj.kubeconfig)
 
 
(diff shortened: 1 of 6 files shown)


kqueen/blueprints/api/views.py
```diff
@@ -156,6 +156,27 @@ def cluster_kubeconfig(pk):
...
+            'progress': 0,
+            'result': obj.get_state()
+        }
+    except Exception:
+        progress = {
+            'response': 500,
+            'progress': 0,
+            'result': config.get('CLUSTER_UNKNOWN_STATE')
+        }
+    return jsonify(progress)
```
tomkukral: Do you really want to return `200` all the times and put return codes into content?

I think you should return 500 in case of failure.

PR #179, opened 2017-12-20
Move update password to dedicated API endpoint

atengler:
Move update password to separate API endpoint


kqueen/blueprints/api/views.py
```diff
@@ -293,6 +321,28 @@ class DeleteUser(DeleteView):
 api.add_url_rule('/users/<uuid:pk>', view_func=DeleteUser.as_view('user_delete'))
 
 
+@api.route('/users/<uuid:pk>/updatepw', methods=['PATCH'])
```
tomkukral: Can you add this to swagger?

PR #182, opened 2017-12-21
Update swagger API definitions

atengler:
Update swagger definitions for both 2.0 (code generation) and 3.0 (swagger UI) standarts to match current state of the API. Use flask-swagger-ui 3.6.0 or newer to support OpenAPI 3.0 on swagger UI page.


tomkukral:
Do you think we still need to have swagger 2.0 as well?

PR #183, opened 2017-12-22
Encryption fix

tomkukral:
Fix edge cases during encryption

### DIFF
--- a/kqueen/storages/etcd.py
+++ b/kqueen/storages/etcd.py
@@ -137,18 +137,16 @@ class Field:
         if not self.encrypted:
             return serialized
 
-        if self.value is None:
-            raise Exception('None value can\'t be encrypted')
+        if self.value is not None:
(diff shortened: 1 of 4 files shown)


kqueen/storages/etcd.py
```diff
@@ -132,11 +132,16 @@ def _unpad(self, s):
     def encrypt(self):
         """Encrypt stored value"""
 
+        serialized = self.serialize()
+
         if not self.encrypted:
-            return self.serialize()
+            return serialized
+
+        if self.value is None:
```
atengler: Isn't it possible to have required=False attribute on model with encrypted=True? This feels like it's going to raise unnecessarily

tomkukral: Good point ... `serialize` returns `None` for missing value and `encrypt` should do it same way.

Issue #190, opened 2018-01-03
Google engine should explain that project ID is needed and not project name


pupapaik:
Google provisioner should explain that project ID is needed and not project name. It confused people, who tries to add project name instead of ID.

tomkukral:
Closed by #191

PR #191, opened 2018-01-03
GCE should explain project ID not project

pupapaik:
Google provisioner should explain that project ID is needed and
not project name.

Closes-bug: #190


tomkukral:
You must use `Closes` not `Closes-bug`. See https://help.github.com/articles/closing-issues-using-keywords/

Issue #192, opened 2018-01-03
Unable to cancel/delete clusters in Deploying status


pupapaik:
Bug description: For unknown reasons my clusters aren't deploying past 1%, and I have no way to cancel/remove them.

Solution:
Most of cloud providers like Azure or GKE do not allow to delete cluster during provisioning, therefore I propose following:
- [ ] Put message that clusters cannot be deleted during provisioning (UI)
- [ ] Put configurable timeout with default value 1 hour then goes to Error (API - #193)
- [ ] Error cluster can be removed, but with confirmation message that user should contact his cloud provider administrator to verify provider side (delete manually if needed) (UI)

Issue #162, opened 2017-12-08
Improve handling of cluster status


atengler:
We should improve handling of cluster status in case of provisioning failure (not always switches the status to error).

epcim:
Ref - #239

PR #193, opened 2018-01-03
Configurable timeout for Provisioner

pupapaik:
This allows to configure timeout during cluster provisioning in case
of unknown errors or network disconnection. By default is it si set
on 1 hour.

Closes: #192
Refs: #162

### DIFF
--- a/kqueen/models.py
+++ b/kqueen/models.py
@@ -53,7 +53,7 @@ class Cluster(Model, metaclass=ModelMeta):
             self.save()
 
         # check for stale clusters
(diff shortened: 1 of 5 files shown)


tomkukral:
note to me: add tests

kqueen/models.py
```diff
@@ -48,6 +49,11 @@ def get_state(self):
             self.state = cluster['state']
             self.save()
 
+        # check for stale clusters
+        max_age = timedelta(hours=config.get('PROVISIONER_TIMEOUT'))
```
tomkukral: This won't work in case of state is returned in L47.

PR #198, opened 2018-01-12
Scaling of cluster nodes

atengler:
This PR contains following changes:

- Fix compatibility with new version of Azure Contrainer Service API. Raise version of `azure-mgmt-containerservice` package to 3.0.0 and implement new way of getting Kubeconfig
- Add `resize` method to base engine class to allow cluster node scaling
- Add `Resizing` state for Cluster object
- Implement resize for GCE and AKS engines
- Add `cluster_resize` API endpoint

### DIFF
--- a/kqueen/blueprints/api/views.py
+++ b/kqueen/blueprints/api/views.py
@@ -179,6 +179,26 @@ def cluster_progress(pk):
     return jsonify(progress)
 
 
+@api.route('/clusters/<uuid:pk>/resize', methods=['PATCH'])
+@jwt_required()
+def cluster_resize(pk):
+    obj = get_object(Cluster, pk, current_identity)
+
+    data = request.json
+    if not isinstance(data, dict) or (isinstance(data, dict) and 'node_count' not in data):
+        abort(400)
+
+    res_status, res_msg = obj.engine.resize(data['node_count'])
+
+    if not res_status:
+        logger.error('Resizing failed: {}'.format(res_msg))
+        abort(500, description=res_msg)
+
+    # get object with updated metadata
+    output = obj.engine.cluster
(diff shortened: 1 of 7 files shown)


kqueen/blueprints/api/views.py
```diff
@@ -179,6 +179,26 @@ def cluster_progress(pk):
...
+@api.route('/clusters/<uuid:pk>/resize', methods=['PATCH'])
+@jwt_required()
+def cluster_resize(pk):
+    obj = get_object(Cluster, pk, current_identity)
+
+    data = request.json
+    if not isinstance(data, dict) or (isinstance(data, dict) and 'node_count' not in data):
+        abort(400)
+
+    res_status, res_msg = obj.engine.resize(data['node_count'])
```
tomkukral: I'd prefer to have `resize` method on cluster instance.

kqueen/engines/aks.py
```diff
@@ -125,47 +160,17 @@ def provision(self, **kwargs):
...
-                            }
-                        ]
-                    }
-                }
-            }
-        }
+        managed_cluster = ManagedCluster(
+            self.location,
+            dns_prefix=self.resource_group_name,
+            kubernetes_version='1.7.7',
```
tomkukral: Please keep `TODO: fix hardcoded params`

kqueen/blueprints/api/views.py
atengler: We currently do it this way on all views, so maybe we can move it at some later date during refactoring

kqueen/engines/aks.py
atengler: OK, will do

PR #206, opened 2018-01-24
don't save kubeconfig to file

tomkukral:
We can pass config dict directly to Kubernetes python client and it will
solve our problems with concurency and cleaning old config files.

### DIFF
--- a/kqueen/kubeapi.py
+++ b/kqueen/kubeapi.py
@@ -1,12 +1,12 @@
 """Kubernetes client wrapper."""
 
 
-from kubernetes import config, client
+from kubernetes import client
+from kubernetes.config.kube_config import KubeConfigLoader
 from kubernetes.client.rest import ApiException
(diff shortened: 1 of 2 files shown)


kqueen/kubeapi.py
```diff
@@ -29,31 +29,34 @@ def __init__(self, **kwargs):
...
 
         Returns:
             str: Kubeconfig file path
 
         """
-        # TODO: make configfile name random
-        configfile = '/tmp/kubernetes'
-        f = open(configfile, 'w')
+
+        # This super-ugly code configuration is cause by wrong desing of config-loading
```
atengler: caused instead of cause, design instead of desing :man_student:

tomkukral: It isn't a grammar but typo!

atengler: Typo nazis then?

PR #212, opened 2018-01-31
Organization management

atengler:
Allow filtering of requested resource lists by namespace and check if its possible to delete organization.


kqueen/blueprints/api/generic_views.py
```diff
@@ -140,15 +148,34 @@ class ListView(GenericView):
     action = 'list'
 
     def set_object(self, *args, **kwargs):
+        if request.args.get('all_namespaces'):
```
tomkukral: Why `args`? I'd expect it to be `kwarg` - `namespace='all'`

kqueen/blueprints/api/generic_views.py
```diff
@@ -140,15 +148,34 @@ class ListView(GenericView):
...
 
     def get_content(self, *args, **kwargs):
+        if request.args.get('all_namespaces'):
+            objs = []
+            for obj in self.obj:
+                namespace = obj._object_namespace
+                obj_dict = obj.get_dict(expand=True)
+                obj_dict['namespace'] = namespace
+                objs.append(obj_dict)
+            return objs
```
tomkukral: We are returning:

* object for single namespace
* list of dicts (not objects) for all_namespaces

It this desired behavior? Why do we don't return list of object?*

* and add _namespace property to objects?

kqueen/models.py
```diff
@@ -376,6 +376,44 @@ class Organization(Model, metaclass=ModelMeta):
     policy = JSONField()
     created_at = DatetimeField(default=datetime.utcnow)
 
+    def is_deletable(self):
+        remaining = []
+        if User.list(self.namespace, return_objects=False):
+            all_users = User.list(None).values()
+            users = [u for u in all_users if u.namespace == self.namespace]
```
tomkukral: not a blocker: I'd prefer to use more lines of code and make this code more readable.

kqueen/models.py
```diff
@@ -376,6 +376,44 @@ class Organization(Model, metaclass=ModelMeta):
...
+                })
+        if remaining:
+            return False, remaining
+        return True, remaining
+
+    def delete(self):
+        deletable, remaining = self.is_deletable()
+        if deletable:
+            return super().delete()
+        resources = ','.join(['{} {}'.format(r['object'], r['uuid']) for r in remaining])
```
tomkukral: This is hard to read, can you make this less magical?

kqueen/blueprints/api/generic_views.py
atengler: List is a GET only endpoint, so the `all_namespaces` param is passed via urlencode and appears as arg on request

kqueen/blueprints/api/generic_views.py
atengler: I agree, that we could prefix the namespace attribute, so it does not look as default part of the schema for given resource, but I don't think whether its object or dict at this point, I think it gets serialized in the exactly same way either way before its returned.

kqueen/models.py
atengler: In this case I can't really see how to make this even more readable :D.

kqueen/models.py
atengler: This one could get more readable, true that. I'll make it so

kqueen/blueprints/api/generic_views.py
tomkukral: Thanks, waiting for fix.

kqueen/models.py
tomkukral: This on is acceptable.

PR #214, opened 2018-02-06
LDAP authentication

tomkukral:
Add LDAPAuth class to provide LDAP authentication.


tomkukral:
Code is working, I just need to properly test this code and cover all added lines.

kqueen/auth/common.py
```diff
@@ -22,14 +35,28 @@ def authenticate(username, password):
...
     users = list(User.list(None, return_objects=True).values())
     username_table = {u.username: u for u in users}
     user = username_table.get(username)
+
     if user:
-        user_password = user.password.encode('utf-8')
         given_password = password.encode('utf-8')
-        if user.active and bcrypt.checkpw(given_password, user_password):
+
+        # setup auth class
```
tomkukral: # default auth class

kqueen/auth/ldap.py
```diff
@@ -0,0 +1,32 @@
...
+
+            return False
+
+        except ldap.LDAPError as e:
+            logger.error(e)
+
+        finally:
+            self.connection.unbind()
+
+        return False
```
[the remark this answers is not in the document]
tomkukral: We don't need user CRUD, LDAP is read-only.

kqueen/auth/local.py
```diff
@@ -0,0 +1,24 @@
...
+        """
+
+        if isinstance(user, User):
+            user_password = user.password.encode('utf-8')
+            given_password = password
+
+            if bcrypt.checkpw(given_password, user_password):
+                return user, None
+
+        return None, "Local authentication failed"
```
naumvd95: logger?

kqueen/auth/test_ldap.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def test_raise_on_missing_uri(self):
+        with pytest.raises(Exception, msg='Parameter uri is required'):
+            LDAPAuth()
+
+    def test_login_pass(self):
+        password = 'heslo123'
+        user, error = self.auth_class.verify(self.user, password)
+
+        assert isinstance(user, User)
```
naumvd95: assertion error descriptions?

kqueen/auth/test_ldap.py
```diff
@@ -0,0 +1,49 @@
...
+        assert error is None
+
+    def test_login_bad_pass(self):
+        password = 'abc'
+        user, error = self.auth_class.verify(self.user, password)
+
+        assert not user
```
naumvd95: assertion error descriptions?

kqueen/auth/test_ldap.py
```diff
@@ -0,0 +1,49 @@
...
+        assert error == "Invalid LDAP credentials"
+
+    def test_bad_server(self):
+        password = 'heslo123'
+        auth_class = LDAPAuth(uri="ldap://127.0.0.1:55555")
+
+        user, error = auth_class.verify(self.user, password)
+        assert not user
```
naumvd95: assertion error descriptions?

kqueen/auth/common.py
```diff
@@ -22,14 +35,30 @@ def authenticate(username, password):
...
+
     if user:
-        user_password = user.password.encode('utf-8')
         given_password = password.encode('utf-8')
-        if user.active and bcrypt.checkpw(given_password, user_password):
-            return user
+
+        # fallback to local auth
+        if not user.auth:
+            user.auth = "LocalAuth"
```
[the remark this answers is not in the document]
tomkukral: This one is default options because we don't have this field yet. I'll add logging for all methods.

kqueen/auth/ldap.py
```diff
@@ -0,0 +1,75 @@
...
+        Returns:
+            dn (str): LDAP dn, like 'cn=admin,dc=example,dc=org
+        """
+
+        segments = []
+
+        if '@' in email:
+            cn, dcs = email.split('@')
+        else:
+            cn = email
```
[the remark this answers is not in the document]
tomkukral: Sorry, I don't understand what do you want to change.

kqueen/auth/ldap.py
```diff
@@ -0,0 +1,75 @@
...
+        if cn:
+            segments.append('cn={}'.format(cn))
+
+        if '.' in dcs:
+            for s in dcs.split('.'):
+                segments.append('dc={}'.format(s))
+
+        return ','.join(segments)
+
+    def verify(self, user, password):
```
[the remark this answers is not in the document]
tomkukral: Sorry, I don't understand ... can you be more specific?

kqueen/auth/test_ldap.py
tomkukral: already on on L26: `assert error is None`

kqueen/auth/ldap.py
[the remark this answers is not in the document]
naumvd95: if @ not in email, we may get 'blabla.org'
maybe need to cut out '.org' in this case?

kqueen/auth/ldap.py
```diff
@@ -0,0 +1,75 @@
...
+class LDAPAuth(BaseAuth):
+    def __init__(self, *args, **kwargs):
+        """
+        Implementation of :func:`~kqueen.auth.base.__init__`
+        """
+
+        super(LDAPAuth, self).__init__(*args, **kwargs)
+
+        if not hasattr(self, 'uri'):
+            raise Exception('Parameter uri is required')
```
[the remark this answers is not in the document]
tomkukral: Good point, I have added try/except to auth function where this function is called.

kqueen/auth/ldap.py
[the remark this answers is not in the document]
tomkukral: if you omit `@` in email then your username is `blabla.org` and this function is working find IMHO.

Can you update tests (https://github.com/Mirantis/kqueen/blob/ldap/kqueen/auth/test_ldap.py#L44) if there is a case which I'm missing?

kqueen/auth/ldap.py
```diff
@@ -0,0 +1,75 @@
...
+        if '.' in dcs:
+            for s in dcs.split('.'):
+                segments.append('dc={}'.format(s))
+
+        return ','.join(segments)
+
+    def verify(self, user, password):
+        """Implementation of :func:`~kqueen.auth.base.__init__`
+
+        This function tries to bind LDAP and returns result
```
katyafervent: Could you add mark about function result. For me the result of logical verify returns nothing and reason of error in case verification is not made. Why do we return tuple with the user as a first item?

kqueen/auth/ldap.py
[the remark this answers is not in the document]
naumvd95: hm. yep , i agree

kqueen/auth/ldap.py
tomkukral: Return values are defined in base class (https://github.com/Mirantis/kqueen/blob/ldap/kqueen/auth/base.py#L22) - is this what you want?

I choosed to return tthe user as well as error mesage to be able to provide more information about process, for example User logged in but password should be change.

Another case can be updating User - like writing some attributes to User or modification of `last_login` field in the future.

tomkukral:
Hi @katyafervent @naumvd95, thanks for your comments!

I tried to address all of them.

kqueen/auth/common.py
```diff
@@ -1,15 +1,28 @@
...
 
+def get_auth_instance(name):
+    config = current_config()
+
+    auth_config = config.get("AUTH", {}).get(name, {})
+
+    module = importlib.import_module('kqueen.auth')
+    auth_class = getattr(module, name)
+
+    return auth_class(**auth_config)
```
atengler: Considering `auth_class` can be None, you could handle the possible `NoneType not callable`  error here

katyafervent:
Looks good to me.

kqueen/models.py
```diff
@@ -428,6 +428,7 @@ class User(Model, metaclass=ModelMeta):
     role = StringField(required=True)
     active = BoolField(required=True)
     metadata = JSONField(required=False)
+    auth = StringField()
```
tomkukral: Add default value

PR #216, opened 2018-02-07
file handlers to root logger, dockerfile refactor:

naumvd95:
- root logger configuration added to server.py as a 'core' level
to avoid possible reinitializing and handler duplicating.
- filehandler configuration added
- kqueen user, gosu service defined to avoid root-actions, provide good dockerfile practice
- refactoring dockerfile actions to increase visibility and provide good practice

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,32 @@
 FROM python:3.6
 
-# prepare directory
+# explicitly set user/group IDs
+RUN groupadd -r -g 999 kqueen && \
+    useradd -r -g kqueen -u 999 -m -s /sbin/nologin -c "Kqueen user" kqueen
+
+COPY . /code
+
 WORKDIR /code
 
-# copy app
-COPY . .
+RUN mkdir -p log/kqueen
 
 # install from local file
 RUN pip install .
 
-# run app
-CMD ./entrypoint.sh
+# grab gosu for easy step-down from root
+ENV GOSU_VERSION 1.7
+RUN set -x \
+    && apt-get update && apt-get install -y --no-install-recommends ca-certificates wget && rm -rf /var/lib/apt/lists/* \
+    && wget -O /usr/local/bin/gosu "https://github.com/tianon/gosu/releases/download/$GOSU_VERSION/gosu-$(dpkg --print-architecture)" \
+    && wget -O /usr/local/bin/gosu.asc "https://github.com/tianon/gosu/releases/download/$GOSU_VERSION/gosu-$(dpkg --print-architecture).asc" \
+    && export GNUPGHOME="$(mktemp -d)" \
+    && gpg --keyserver ha.pool.sks-keyservers.net --recv-keys B42F6819007F00F88E364FD4036A9C25BF357DD4 \
+    && gpg --batch --verify /usr/local/bin/gosu.asc /usr/local/bin/gosu \
+    && rm -r "$GNUPGHOME" /usr/local/bin/gosu.asc \
+    && chmod +x /usr/local/bin/gosu \
+    && gosu nobody true \
+    && apt-get purge -y --auto-remove ca-certificates wget
+
+ENTRYPOINT ["./entrypoint.sh"]
+
+CMD gunicorn --config kqueen/gunicorn.py kqueen.wsgi
--- a/docker-compose.demo.yml
+++ b/docker-compose.demo.yml
@@ -2,6 +2,8 @@ version: '2'
 services:
   api:
     image: kqueen/api:v0.16
+    volumes:
+      - /tmp/kqueen:/code/log/kqueen
     ports:
       - 127.0.0.1:5000:5000
     depends_on:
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -1,13 +1,32 @@
...
-# copy app
-COPY . .
+RUN mkdir -p log/kqueen
 
 # install from local file
 RUN pip install .
 
-# run app
-CMD ./entrypoint.sh
+# grab gosu for easy step-down from root
```
tomkukral: Why? You can use `USER` command IMO.

Dockerfile
```diff
@@ -1,13 +1,32 @@
...
+RUN groupadd -r -g 999 kqueen && \
+    useradd -r -g kqueen -u 999 -m -s /sbin/nologin -c "Kqueen user" kqueen
+
+COPY . /code
+
 WORKDIR /code
 
-# copy app
-COPY . .
+RUN mkdir -p log/kqueen
```
tomkukral: You should expect container to be running on read-only mode so please move logs out of source code.

docker-compose.demo.yml
```diff
@@ -2,6 +2,8 @@ version: '2'
 services:
   api:
     image: kqueen/api:v0.16
+    volumes:
```
tomkukral: Why do you need to mount it in demo environment? You can take a look into containers for demo environent.

It won't provide persistent storage in any way.

entrypoint.sh
```diff
@@ -11,4 +11,8 @@ if [[ "$BOOTSTRAP_ADMIN" > 0 ]] ; then
     python bootstrap_admin.py ${BOOTSTRAP_ADMIN_ORGANIZATION} ${BOOTSTRAP_ADMIN_NAMESPACE} ${BOOTSTRAP_ADMIN_USERNAME} ${BOOTSTRAP_ADMIN_PASSWORD}
 fi
 
-exec gunicorn --config kqueen/gunicorn.py kqueen.wsgi
+# Setup permissions
+chown -R kqueen:kqueen /tmp /code /var/log/kqueen 2>/dev/null || :
+chmod -R 700 /tmp /code /var/log/kqueen 2>/dev/null || :
+
+exec gosu kqueen "$@"
```
tomkukral: Why are you droping permissions here? I'd expect to drop it before starting entrypoint.

kqueen/config/demo.py
```diff
@@ -4,6 +4,8 @@
 class Config(BaseConfig):
     DEBUG = False
     LOG_LEVEL = 'INFO'
+    LOG_PATH = 'log/kqueen'
```
tomkukral: Please move logs out of source code.

kqueen/server.py
```diff
@@ -14,7 +14,22 @@
...
 
+#Logging Configuration
+config = current_config(config_file=None)
+config.setup_policies()
+
+log_path = config.get('LOG_PATH')
+log_file = config.get('LOG_FILE')
+
+logging.basicConfig(level=logging.DEBUG,
+                    format='%(asctime)s %(name)-12s %(levelname)-8s %(message)s',
```
tomkukral: please make format configurable.

Dockerfile
naumvd95: official docker best practises provide several pros of gosu using, pls take a look https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

Dockerfile
naumvd95: imo there are 3 ways to  provide logging directory:
1) in docker file
2) in entry point
3) in python backend
which one do you prefer? or you have any another proposals?

docker-compose.demo.yml
naumvd95: ok, btw, do we need to map logging into etcd storage?

entrypoint.sh
naumvd95: imo there are 2 ways to do it:
1) drop in docker file (imo its not beauty for file-reference)
2) create separate file, called like 'setup.sh' and paste all config there  and run it from entrypoint.

kqueen/server.py
naumvd95: imo there are 2 ways:
1) make config.ini file for pretty configurations
2) provide additional variables ink queen config and link it here
which one do you prefer?

kqueen/config/demo.py
naumvd95: why? if i remove it from all configs, api should not start. If we decided to provide logging in files, we should provide default path's , otherwise we should provide some kind of additional checker 'FILE_LOGGING_ON/OFF' and wrap all stuff into it

Dockerfile
tomkukral: If you need `sudo` ...

We don't need sudo IMHO.

Dockerfile
tomkukral: If fine with providing loggin directory in this place but please don't log into same directory as source code is living.

kqueen/config/demo.py
tomkukral: I just wanted to log into different directory, something like "/var/log/queen/ or anything else.

naumvd95: oh, got it

Dockerfile
naumvd95: yep, you are right, ill remove it

PR #217, opened 2018-02-09
jenkins engine status checker changes

naumvd95:
previous was useless, it wasnot failed due incorrect creds.

PAY  ATTENTION , that this change provide red status of provisioner on UI after creation and backend tracing,
but we already have flush message 'Provisioner successfully created' on main board.

TODO: refactor api-requests realated to provisioner creating , and add engine_status checker into it


kqueen/engines/gce.py
```diff
@@ -286,13 +286,44 @@ def cluster_get(self):
...
         if response.status_code == 401:
+            try:
+                self.cluster_list()
+            except Exception as e:
+                msg = 'Failed to discover GCE project. Check that credentials is valid. Error:'
+                logger.exception(msg)
+                return config.get('PROVISIONER_UNKNOWN_STATE')
+            clusters = self.cluster_list()
+            if len(clusters) != 0:
+                logger.critical('Additional clusters discovered. \
```
atengler: I think this should be `warning` level log, it doesn't bother us too much, that there are other clusters in the project

atengler:
Regarding the message in UI, `Provisioner successfully created` refers to object being successfully created, which is true even though the backend is unreachable for whatever reason. The message changes only when backend returns error and the Provisioner object is not created at all, but this can be changed in the future.

kqueen/engines/gce.py
atengler: You can instantiate engine by doing:
cluster = Cluster() # there may be some mandatory args
engine = engine_class(cluster)

PR #225, opened 2018-02-14
refactor logging arch for api

naumvd95:
- generic logging arch provided both for api and ui parts (take a look on ui part  - https://github.com/Mirantis/kqueen-ui/pull/18)
- refactor pip packanges installing (pip install . does not install all packages, proofs:
https://pastebin.com/ZCvxkmZK (with -r requirements)
https://pastebin.com/kWVeGbzu ( without)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,6 @@ WORKDIR /code
 
 # copy app
 COPY . .
-
-# install from local file
 RUN pip install .
 
 # run app
--- a/kqueen/auth.py
+++ b/kqueen/auth.py
@@ -131,7 +131,7 @@ def is_authorized(_user, policy_value, resource=None):
     try:
         condition = conditions[policy_value]
     except KeyError:
-        logger.error('Policy evaluation failed. Invalid rule: {}'.format(str(policy_value)))
+        logger.exception('Policy evaluation failed. Invalid rule: {}'.format(str(policy_value)))
         return False
 
     if ROLE == 'superadmin':
@@ -145,7 +145,7 @@ def is_authorized(_user, policy_value, resource=None):
             logger.error('Policy evaluation did not return boolean: {}'.format(str(authorized)))
             authorized = False
     except Exception as e:
-        logger.error('Policy evaluation failed: {}'.format(repr(e)))
+        logger.exception('Policy evaluation failed: ')
         authorized = False
     logger.debug('User {} id {} authorized as {}'.format(user['username'], user['id'], user['role']))
     return authorized
--- a/kqueen/auth.py
+++ b/kqueen/auth.py
@@ -136,7 +136,7 @@ def is_authorized(_user, policy_value, resource=None):
 
     if ROLE == 'superadmin':
         # no point in checking anything here
-        logger.debug('User [] id [] authorized as {}'.format(user['username'], user['id'], user['role']))
+        logger.debug('User {} id {} authorized as {}'.format(user['username'], user['id'], user['role']))
         return True
 
     try:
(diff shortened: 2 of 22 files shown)


Dockerfile
```diff
@@ -5,9 +5,7 @@ WORKDIR /code
 
 # copy app
 COPY . .
-
-# install from local file
-RUN pip install .
+RUN pip install . && pip install -r requirements.txt
```
tomkukral: Why? We don't want to install all testing and support requirements, thanks why we are installing from `setup.py`.

These packages aren't requred for runtime:
```
    'coveralls',
    'faker',
    'flake8',
    'pytest',
    'pytest-cov',
    'pytest-env',
    'pytest-flask',
    'pytest-ordering',
```

tomkukral: BTW, `requirements.txt` isn't used anywhere. We keep it just be backward compatible and easier to onboard for new developers.

kqueen/blueprints/api/views.py
```diff
@@ -25,7 +25,7 @@
 import yaml
 
 config = current_config()
-logger = logging.getLogger(__name__)
+logger = logging.getLogger('kqueen_api')
```
tomkukral: Why are you hidding and ignoring python logger hierarchy? We can use root logger and use propagation to catch all logs.

kqueen/server.py
```diff
@@ -10,11 +10,14 @@
...
 from flask_jwt import JWT
 from flask_swagger_ui import get_swaggerui_blueprint
+from kqueen.utils.loggers import setup_logging
 from werkzeug.contrib.cache import SimpleCache
 
 import logging
 
-logger = logging.getLogger(__name__)
+# Logging configuration
+setup_logging('/code/kqueen/utils/logger_config.yml')
```
tomkukral: How is this path going to be created in on machine? This will work only in container.

kqueen/utils/loggers.py
```diff
@@ -0,0 +1,24 @@
+import os
+import yaml
+import logging.config
+import logging
+
+
+def setup_logging(default_path='/code/kqueen_ui/utils/logger_config.yml', default_level=logging.INFO, env_key='LOG_CFG'):
```
tomkukral: Remove hardcoded default path or put them to reasonable path.

tomkukral: Please don't use `LOG_CFG`, at least prefix it with `KQUEEN_`.

Or better use configuration via `kqueen.config`.

Dockerfile
naumvd95: ok, ill remove it

kqueen/blueprints/api/views.py
naumvd95: __name__  var used only for module naming, i already defined *module* naming in logger formatting (logger_config.py), it will be good for:
1) reuse 1 logger and does not create multiple loggers -> increase performance
2) all logging arch stored in one file -> increase visibility

so in fact , no cons, just pros

kqueen/server.py
naumvd95: yep, i already started to wrap it on  kqueen config-vars

kqueen/utils/loggers.py
naumvd95: ok, got it

kqueen/blueprints/api/views.py
naumvd95: and its also needed in case of splitting user logging and common logging, but needed to provide common arch for api and ui parts , so please take a look on ui part

tomkukral: Ok, thanks for clarification, your solution is easier and simple, let's use it.

kqueen/auth.py
```diff
@@ -136,6 +136,7 @@ def is_authorized(_user, policy_value, resource=None):
 
     if ROLE == 'superadmin':
         # no point in checking anything here
+        logger.debug('User [] id [] authorized as {}'.format(user['username'], user['id'], user['role']))
```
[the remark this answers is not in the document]
naumvd95: oh) lol ) i can say many abstruse things about that.... but in fact, its just a mistake =))) sorry, ill fix it

atengler:
Could you maybe replace logger.error with logger.exception in API as well? To get the tracebacks from handled API errors

naumvd95:
pls re-review

PR #230, opened 2018-02-27
test cyclomatic completixy

tomkukral:
Run `mccabe` to check cyclomatic complexity.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ jobs:
     - pip install -e ".[test]"
     - pip check
     script:
-    - python3 -m flake8
+    - python3 -m flake8 --max-complexity 10
   - stage: test
     before_install:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     - pip install -e ".[test]"
     - pip check
     script:
-    - python3 -m flake8
+    - python3 -m flake8 --max-complexity 10
```
naumvd95: we can increase amount if necessary ) just for notice
```This practice was adopted by the NIST Structured Testing methodology, with an observation that since McCabe's original publication, the figure of 10 had received substantial corroborating evidence, but that in some circumstances it may be appropriate to relax the restriction and permit modules with a complexity as high as 15.```

tomkukral: Most of the code can by easily refactored ... except `Cluster.topology_data` which is a nightmare.

PR #233, opened 2018-02-28
    GCE engine_status, cluster _list methods implemented

naumvd95:


kqueen/engines/gce.py
```diff
@@ -284,7 +284,38 @@ def cluster_get(self):
...
+            for cluster in clusters:
+                state = STATE_MAP.get(cluster['status'], config.get('CLUSTER_UNKNOWN_STATE'))
+                key = 'cluster-{}-{}'.format(cluster['name'], self.cluster_id or None)
+                item = {
+                    'key': key,
+                    'name': self.cluster_id or cluster['name'],
+                    'id': self.cluster.id or None,
+                    'state': state,
+                    'metadata': {
+                        'Node config': cluster['nodeConfig'],
```
atengler: Make the dict keys slugs pls, `node_config` instead of `Node Config` and `master_version` or `current_master_version` instead of `Current Master Version`

naumvd95: ok

atengler:
You need to add `**kwargs` to `engine_status` methods on base and manual provisioners as well for everything to work properly

naumvd95:
added

katyafervent:
Please, change the commit message according to https://github.com/erlang/otp/wiki/writing-good-commit-messages and http://who-t.blogspot.ru/2009/12/on-commit-messages.html

naumvd95:
@katyafervent its already done, take a look on commit messages

PR #234, opened 2018-02-28
Update devenv script

katyafervent:
Add metadata attributes to mock cluster objects and parameter in provisioner objects.
After that all mock objects created with devenv script are available in UI.


tomkukral:
These `paremeters` and `metadata` properties are optional ... why do we require them in UI? 

cc @atengler

atengler:
Existence of attribute defined on model is never optional, it just means they can be empty. I can't really write a code in a way that attributes on stuff may or may not be all the time, that bloats the app big time.

PR #235, opened 2018-02-28
gce engine check refactoring

naumvd95:
provisioner statuses:
OK- url access, auth is success
UNKNOWN- url access, auth is failed
ERROR- url unavailable

- cluster listing check implemented in engine_status for future deatures, like 'auto attaching existing clusters into kqueen'
- jenkins engine status refactored to get kwargs instead of config ( if exist)
- added zone defaults '-' (this symbol means, that method should work for all zones)

p.p.p.s need to merge this request into gce-list brunch
then update pr https://github.com/Mirantis/kqueen/pull/233/files
and merge it to master

### DIFF
--- a/kqueen/engines/gce.py
+++ b/kqueen/engines/gce.py
@@ -117,7 +117,7 @@ class GceEngine(BaseEngine):
         # Client initialization
         self.service_account_info = kwargs.get('service_account_info', {})
         self.project = kwargs.get('project', '')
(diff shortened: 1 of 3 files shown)


atengler:
I think you have the necessary rights to merge stuff between your development branches, is it necessary to do this through PR?

naumvd95:
ok, ill do that,  i prefer this structure just to increase visibility

PR #237, opened 2018-03-01
Provisioner alive check

atengler:
Similar to ListCluster, check status of Provisioner on list request, so we can disallow cluster provisioning with unhealthy Provisioner object.

### DIFF
--- a/kqueen/blueprints/api/views.py
+++ b/kqueen/blueprints/api/views.py
@@ -216,7 +216,7 @@ class ListProvisioners(ListView):
 
     def get_content(self, *args, **kwargs):
         provisioners = self.obj
-        if config.get('PROVISIONER_STATE_ON_LIST', True):
+        if config.get('PROVISIONER_STATE_ON_LIST'):
             try:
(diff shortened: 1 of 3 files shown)


kqueen/blueprints/api/views.py
```diff
@@ -210,6 +210,34 @@ def cluster_resize(pk):
...
+                    loop = asyncio.new_event_loop()
+                    asyncio.set_event_loop(loop)
+                # run coroutines and close loop
+                loop.run_until_complete(asyncio.gather(*[self._update_provisioner(p) for p in provisioners]))
+                loop.close()
+            except Exception as e:
+                logger.exception('Asyncio loop is NOT available, fallback to simple looping: ')
+
+                for p in provisioners:
+                    p.engine_status()
```
katyafervent: how do we check that provisioner is ok? with no exception?

atengler: By calling `engine_status` method, it checks and updates states of Provisioner objects

PR #247, opened 2018-03-10
refactor kqueen metrics mechanism

naumvd95:
previously i cant get users/orgs metrics with:
'''localhost:5000/metrics/'''

- kqueen metrics updating
- logging for above operation (info level)
- prevent overriding env variables for pytest to avoid issues related to directory doesnot exist:
```FileNotFoundError: [Errno 2] No such file or directory: '/tmp/prometheus_client/gauge_all_5850.db'```
    it happens when setup.cfg overrides env var for pytest run

Notes:
We export all application metrics on /metrics and this metric can scraped by any external Prometheus instance. There two ways to authenticate to get this endpoint:

1. Provided valid token in header

TOKEN=$(curl -s -H "Content-Type: application/json" --data '{"username":"admin","password":"default"}' -X POST localhost:5000/api/v1/auth | jq -r '.access_token'); echo $TOKEN; curl -H "Authorization: Bearer $TOKEN"
And then go to localhost:5000/metrics

2. Add scraper IP address to PROMETHEUS_WHITELIST configuration and then

Go to {prometheus_ip}:{port}/metrics

review-docs: https://docs.google.com/document/d/1Bnor6D78fVzOndie52wFHVVbgV0daTOHrVP1IgKZvic/edit#

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,7 +5,7 @@ test=pytest
 python_files = tests.py test_*.py tests_*.py *_tests.py
 env =
 	KQUEEN_CONFIG_FILE=config/test.py
-	prometheus_multiproc_dir=/tmp/prometheus_client/
+	D:prometheus_multiproc_dir=/tmp/prometheus_client/
 
 [flake8]
 ignore = E501
--- a/kqueen/blueprints/metrics/views.py
+++ b/kqueen/blueprints/metrics/views.py
@@ -5,12 +5,15 @@ from flask import request
 from flask_jwt import _jwt_required
 from ipaddress import ip_address
 from ipaddress import ip_network
-from prometheus_client import CollectorRegistry
+from kqueen.blueprints.metrics.helpers import MetricUpdater
 from prometheus_client import CONTENT_TYPE_LATEST
+from prometheus_client import CollectorRegistry
 from prometheus_client import generate_latest
 from prometheus_client import multiprocess
+import logging
 
 metrics = Blueprint('metrics', __name__)
+logger = logging.getLogger('kqueen_api')
 
 
 @metrics.route('/')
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: kqueen/config/demo.py [hunk 1]
```diff
@@ -3,7 +3,7 @@
 
 class Config(BaseConfig):
     DEBUG = False
-    LOG_LEVEL = 'INFO'
+    LOG_LEVEL = 'DEBUG'
```
katyafervent: Log level is set in logging config. This option is unnecessary.

katyafervent: Also, changes in your commit are not logically connected, so it may be better to separate ones anyway

naumvd95: yep, sure

naumvd95:
@atengler @tomkukral seems i had some issues with tests) can u take a look also?

setup.cfg
```diff
@@ -5,7 +5,7 @@ test=pytest
 python_files = tests.py test_*.py tests_*.py *_tests.py
 env =
 	KQUEEN_CONFIG_FILE=config/test.py
-	prometheus_multiproc_dir=/tmp/prometheus_client/
+	D:prometheus_multiproc_dir=/tmp/prometheus_client/
```
atengler: What is this?

naumvd95: prevent overriding env variables for pytest to avoid issues related to directory doesnot exist:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/prometheus_client/gauge_all_5850.db'
it happens when setup.cfg overrides env var for pytest run

naumvd95: for ex https://travis-ci.org/Mirantis/kqueen/jobs/352239789
@atengler

PR #249, opened 2018-03-12
Update logging settings

katyafervent:
* Remove LOG_LEVEL set up from the main config

    Currently, logging config is used to set up log level for each logger.
    So there is no need to provide default log level. If config provided,
    it will be ignored anyway and leads to misunderstandings.
    So now all log level set up are made in a separate logging config file.

    * Change default general_file_handler  log level

    Also, log level of console logger is changed to INFO instead of log level of
    corresponding handler. It was made, because handler has a higher priority
    and if user will change log level to debug, it still remain INFO or user
     have to change both places, which is inconvenient.

    * If Debug is True in config file, set debug log level automatically

      Otherwise keep log level from config

    * Move default logging config path from setup_logging definition to its call

    Otherwise default value will never applied, cause None will be used

### DIFF
--- a/kqueen/config/demo.py
+++ b/kqueen/config/demo.py
@@ -3,7 +3,6 @@ from .base import BaseConfig
 
 class Config(BaseConfig):
     DEBUG = False
-    LOG_LEVEL = 'DEBUG'
     LOG_CONFIG = 'kqueen/utils/logger_config.yml'
 
     KQUEEN_HOST = '0.0.0.0'
--- a/kqueen/utils/loggers.py
+++ b/kqueen/utils/loggers.py
@@ -4,12 +4,21 @@ import logging.config
 import logging
 
 
-def setup_logging(path='kqueen/utils/logger_config.yml', default_level=logging.INFO):
+def setup_logging(path, debug_mode):
+    default_level = 'INFO'
 
     if os.path.exists(path):
         with open(path, 'rt') as f:
             try:
                 config = yaml.safe_load(f.read())
+                loggers = config['loggers']
+                for logger, value in loggers.items():
+                    if debug_mode:
+                        loggers[logger]['level'] = 'DEBUG'
+                    else:
+                        current_level = value.get('level')
+                        loggers[logger]['level'] = default_level if not current_level else current_level
+
                 logging.config.dictConfig(config)
             except Exception as e:
                 print(e)
--- a/kqueen/config/dev.py
+++ b/kqueen/config/dev.py
@@ -3,7 +3,6 @@ from .base import BaseConfig
 
 class Config(BaseConfig):
     DEBUG = True
-    LOG_LEVEL = 'INFO'
     LOG_CONFIG = 'kqueen/utils/logger_config.yml'
 
     # App secret
--- a/kqueen/config/prod.py
+++ b/kqueen/config/prod.py
@@ -3,7 +3,6 @@ from .base import BaseConfig
 
 class Config(BaseConfig):
     DEBUG = False
-    LOG_LEVEL = 'DEBUG'
     LOG_CONFIG = 'kqueen/utils/logger_config.yml'
 
     KQUEEN_HOST = '0.0.0.0'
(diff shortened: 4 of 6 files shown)


kqueen/utils/logger_config.yml
```diff
@@ -43,3 +51,8 @@ loggers:
         level: INFO
         handlers: [error_file_handler, general_file_handler]
         propogate: yes
+
```
naumvd95: remove space

kqueen/utils/logger_config.yml
```diff
@@ -43,3 +51,8 @@ loggers:
...
+    werkzeug:
```
naumvd95: werkzeug logger does not used anywhere, where do u prefer to summon it?

kqueen/config/demo.py
```diff
@@ -3,7 +3,6 @@
 
 class Config(BaseConfig):
     DEBUG = False
-    LOG_LEVEL = 'DEBUG'
     LOG_CONFIG = 'kqueen/utils/logger_config.yml'
```
tomkukral: Can you delete default `LOG_CONFIG` here or don't set in in `server.py` code. 

Please keep default on single place.

kqueen/utils/loggers.py
```diff
@@ -4,12 +4,22 @@
...
         with open(path, 'rt') as f:
             try:
                 config = yaml.safe_load(f.read())
+                loggers = config['loggers']
+                for logger, value in loggers.items():
+                    if debug_mode:
+                        loggers[logger]['level'] = 'DEBUG'
+                    else:
+                        current_level = value.get('level')
+                        is_level_not_set = value['level'] == 0 or current_level is None
```
tomkukral: Can you please refactor this two lines to be more readable. I understand this code but it's hard to read it on the first look.

However, this isn't a blocker.

kqueen/config/demo.py
katyafervent: project config doesn't set default values, it just keep all set up values in one place. But if you want, I'll delete LOG_CONFIG from all the configs

kqueen/utils/loggers.py
katyafervent: Next time advice me how I can refactor a piece of code that you don't like

kqueen/utils/logger_config.yml
katyafervent: It is internal logger and it can be useful to track requests to the server. It looks like 

2018-03-14 10:51:40,478 - _internal - werkzeug - INFO: 127.0.0.1 - - [14/Mar/2018 10:51:40] "GET /ui/login HTTP/1.1" 200 -
2018-03-14 10:51:40,685 - _internal - werkzeug - INFO: 127.0.0.1 - - [14/Mar/2018 10:51:40] "GET /static/favicon/favicon-32x32.png?v1.0=3eepn6WlLO HTTP/1.1" 200 -
2018-03-14 10:51:42,380 - _internal - werkzeug - INFO: 127.0.0.1 - - [14/Mar/2018 10:51:42] "POST /ui/login HTTP/1.1" 302 -

it is defined here https://github.com/pallets/werkzeug/blob/60670cb98db1934b7aa7df454cbe4508c467f403/werkzeug/_internal.py#L112

kqueen/config/demo.py
katyafervent: I decided to remove hard coded value from, so it is parameter is required to be provided in a config file, otherwise error message about logging in set up correctly will be appear.

kqueen/utils/logger_config.yml
```diff
@@ -33,6 +33,14 @@ handlers:
         backupCount: 20
         encoding: utf8
 
+    flask_file_handler:
+        class: logging.handlers.RotatingFileHandler
+        level: DEBUG
+        formatter: standard
+        filename: /tmp/kqueen_flask.log
```
naumvd95: Pay attention, than in future we need to change dirs to /var/log/kqueen-api
due my changes

PR #252, opened 2018-03-15
docker deployment refactoring

naumvd95:
- volumes for mail server is redundant
- 1/0 values for bool variables turned into True/False to increase visibility
- env variable for DEBUG mode added
- log directories naming refactores to avoid misunderstandings in case of using shared glusterfs volumes

PAY ATTENTION (For Mirantis Only):
This changes dependen on MCP reclas model , so pls review also:
https://gerrit.mcp.mirantis.net/#/q/topic:kqueen+(status:open+OR+status:merged)


naumvd95:
@atengler when we should merge it?

PR #254, opened 2018-03-21
Add Jenkins job URL to cluster metadata

katyafervent:
* Generate Jenkins job url to get quick access to the job in case of fail

### DIFF
--- a/kqueen/engines/jenkins.py
+++ b/kqueen/engines/jenkins.py
@@ -150,11 +150,7 @@ class JenkinsEngine(BaseEngine):
         cluster_build_number = self._get_build_number()
         if not cluster_build_number:
             return {}
-        kubeconfig_url = '{jenkins_url}/job/{job_name}/{build_id}/artifact/kubeconfig'.format(
-            jenkins_url=self.jenkins_url,
-            job_name=self.provision_job_name,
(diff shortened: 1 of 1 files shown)


kqueen/engines/jenkins.py
```diff
@@ -150,11 +150,7 @@ def get_kubeconfig(self):
         cluster_build_number = self._get_build_number()
         if not cluster_build_number:
             return {}
-        kubeconfig_url = '{jenkins_url}/job/{job_name}/{build_id}/artifact/kubeconfig'.format(
-            jenkins_url=self.jenkins_url,
-            job_name=self.provision_job_name,
-            build_id=str(cluster_build_number),
-        )
+        kubeconfig_url = '{job_url}/artifact/kubeconfig'.format(job_url=self._get_jenkins_job_url(str(cluster_build_number)))
```
atengler: You should stringify build number inside `_get_jenkins_job_url` method instead. Otherwise if you feel like this makes the code more readable, I guess its OK

katyafervent: actually, format does not require stringify at all :) So I'll remove it https://docs.python.org/3.1/library/string.html#format-examples

PR #256, opened 2018-03-24
Extend policy on user deletion

katyafervent:
Before admin user was able to delete superadmin.
This change extends policy settings, so permissions
now it is defined for each user role.

### DIFF
--- a/kqueen/blueprints/api/generic_views.py
+++ b/kqueen/blueprints/api/generic_views.py
@@ -8,6 +8,10 @@ from kqueen.auth import is_authorized
 from kqueen.models import Organization
 from .helpers import get_object
 
+import logging
+
+logger = logging.getLogger('kqueen_api')
+
 
 class GenericView(View):
     obj = None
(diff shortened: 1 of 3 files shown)


atengler:
Definitely good point, I didn't notice this flaw. It may be better to handle the condition inside the policy checking method though, can you check this?

kqueen/blueprints/api/generic_views.py
```diff
@@ -47,6 +47,8 @@ def check_authorization(self):
 
         # form policy key
         policy = '{}:{}'.format(_class.__name__.lower(), self.action)
+        if policy == 'user:delete':
+            policy = '{}_{}'.format(policy, self.obj.role)
```
naumvd95: IMO need to get rid of this 'if' case:
1) wrap it in dict.get(,None)
2) or move it into auth view

kqueen/blueprints/api/generic_views.py
```diff
@@ -111,6 +125,13 @@ def dispatch_request(self, *args, **kwargs):
 
         return jsonify({'id': self.obj.id, 'state': 'deleted'})
 
+    def get_policy_key(self):
```
atengler: I would move this method override here - https://github.com/Mirantis/kqueen/blob/9e6955ab850cc7a9bd5b90e8a1932d01fe5420e1/kqueen/blueprints/api/views.py#L414 - you can then skip the if condition, because it will always be 'user:delete'

PR #257, opened 2018-03-27
Set AKS provisioner status on creation

katyafervent:
Before AKS provisioner has Ok status, and error occured only after cluster deployment.
Now valid status is updated on provisioner creation.

### DIFF
--- a/kqueen/engines/aks.py
+++ b/kqueen/engines/aks.py
@@ -2,8 +2,10 @@ from kqueen.config import current_config
 from kqueen.engines.base import BaseEngine
 
 from azure.common.credentials import ServicePrincipalCredentials
+from azure.common.exceptions import AuthenticationError
 from azure.mgmt.containerservice import ContainerServiceClient
 from azure.mgmt.containerservice.models import ManagedCluster
+from msrestazure.azure_exceptions import CloudError
 
 import copy
 import logging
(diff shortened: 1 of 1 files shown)


kqueen/engines/aks.py
```diff
@@ -281,7 +283,30 @@ def cluster_get(self):
...
+            logger.warning('{} Azure Provisioner validation failed with the following reason {}'.format(cls.name,
+                                                                                                        e.message))
+            return config.get('PROVISIONER_UNKNOWN_STATE')
+        client = ContainerServiceClient(credentials, kwargs.get('subscription_id'))
+        try:
+            list(client.managed_clusters.list_by_resource_group(kwargs.get('resource_group_name')))
+        except CloudError as e:
+            logger.warning('Invalid parameters for {} Azure Provisioner: {}'.format(cls.name, e.message))
+            return config.get('PROVISIONER_ERROR_STATE')
+        except Exception as e:
```
naumvd95: take a look on kqueen/engines/gce and do the similar for logging and status returning

naumvd95: for ex:
```            except Exception as e:
                     msg = 'Failed to discover GCE project. Check that credentials is valid. Error:'
                     logger.exception(msg)
                     return config.get('PROVISIONER_UNKNOWN_STATE')```

naumvd95: main goal - provide all engine status checks with equal arch

PR #259, opened 2018-03-28
Docker compose files fix

naumvd95:
- Quotes to avoid composer reference traceback
- Create log-dir for travis tests

### DIFF
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -9,12 +9,12 @@ services:
     restart: always
     environment:
       KQUEEN_CONFIG_FILE: config/prod.py
-      KQUEEN_DEBUG: False
+      KQUEEN_DEBUG: 'False'
       KQUEEN_ETCD_HOST: etcd
       KQUEEN_PROMETHEUS_WHITELIST: '172.16.238.0/24'
       # TODO: set SECRET_KEY
       KQUEEN_SECRET_KEY: ''
-      BOOTSTRAP_ADMIN: True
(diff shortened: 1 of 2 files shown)


naumvd95:
@katyafervent done

docker-compose.production.yml
```diff
@@ -52,7 +52,8 @@ services:
     environment:
       KQUEENUI_PREFERRED_URL_SCHEME: https
       KQUEEN_UI_CONFIG_FILE: config/prod.py
-      KQUEENUI_DEBUG: False
+      KQUEENUI_DEBUG: 'False'
+      # TODO: set SECRET_KEY
       KQUEENUI_SECRET_KEY: 'SecretSecretSecret123'
```
atengler: Both secret keys in this file should be empty strings I default imho. When you don't set them in the specific deployment, application will fail to start which is better than starting with insecure SECRET_KEY, because someone forgot to override this variable.

naumvd95: good point, ill change it

katyafervent: Guys, we should check that key is set before the deployment, at least we can send warning log about that. What do you think?

naumvd95: in case of empty key deployment will fail with verbose trace, u can check it , imo it looks good
@katyafervent

katyafervent: I understand, but how many steps you should done before you got message about secret key is unset

PR #260, opened 2018-03-29
OPS LDAP support, refactoring

naumvd95:
OPS LDAP support, refactoring

    - refactor env-vars arch to provide configuration using only docker env-vars
    - remove apply_env changes due issues with bad overriding in case of multiple value-entries in config file
    - logging increased

### DIFF
--- a/kqueen/auth/ldap.py
+++ b/kqueen/auth/ldap.py
@@ -1,9 +1,11 @@
 from .base import BaseAuth
 
+from kqueen.config import current_config
 import ldap
 import logging
 
 logger = logging.getLogger('kqueen_api')
+config = current_config()
 
 
 class LDAPAuth(BaseAuth):
@@ -17,6 +19,11 @@ class LDAPAuth(BaseAuth):
         if not hasattr(self, 'uri'):
             raise Exception('Parameter uri is required')
 
+        try:
+            ldap.initialize(self.uri)
+        except Exception:
+            logger.exception('Failed to initialize LDAP connection')
+
         self.connection = ldap.initialize(self.uri)
 
     @staticmethod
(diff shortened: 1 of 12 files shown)


kqueen/auth/ldap.py
```diff
@@ -17,6 +19,11 @@ def __init__(self, *args, **kwargs):
         if not hasattr(self, 'uri'):
             raise Exception('Parameter uri is required')
 
+        try:
+            ldap.initialize(self.uri)
+        except Exception:
+            logger.exception('Failed to initialize LDAP connection')
```
katyafervent: you catch and log an exception, but the program flow go further, I suggest to reraise exception or not to catch it HERE at all

kqueen/config/base.py
```diff
@@ -49,6 +49,12 @@ class BaseConfig:
     # Auth settings
     AUTH = {}
```
katyafervent: do we need that one now?

kqueen/config/dev.py
```diff
@@ -5,8 +5,8 @@ class Config(BaseConfig):
     DEBUG = True
     LOG_CONFIG = 'kqueen/utils/logger_config.yml'
 
-    # App secret
-    SECRET_KEY = 'SecretSecretSecret123'
+    # App secret - set this to random string >= 16 chars
+    SECRET_KEY = 'secret123secret123secret123'
```
katyafervent: LL not connected with ldap, extract to another commit please

kqueen/config/utils.py
```diff
@@ -1,9 +1,11 @@
+from distutils import util
```
katyafervent: This also should be extracted

PR #262, opened 2018-04-04
Kqueen ldap

naumvd95:
- refactor env-vars arch to provide configuration using only docker env-vars
- remove apply_env changes due issues with bad overriding in case of multiple value-entries in config file
- logging increased
- splitted on separate commits to increase visibility

p.s. need to refactor tests, related to config-file defining

### DIFF
--- a/kqueen/auth/common.py
+++ b/kqueen/auth/common.py
@@ -41,6 +41,8 @@ def get_auth_instance(name):
 
     module = importlib.import_module('kqueen.auth')
     auth_engine = auth_config.get('engine')
+    logger.debug("Using {} Authentication Engine".format(auth_engine))
+
     if not auth_engine:
         raise Exception('Authentication type is set to {}, but engine class name is not found. '
                         'Please, set it with the "engine" key'.format(name))
@@ -82,6 +84,7 @@ def authenticate(username, password):
             verified_user, verification_error = None, str(e)
 
         if isinstance(verified_user, User) and verified_user.active:
+            logger.info("User {user} passed {method} auth successfully".format(user=user, method=user.auth))
             return verified_user
         else:
             logger.info("User {user} failed auth using {method} auth method with error {error}".format(
--- a/kqueen/auth/common.py
+++ b/kqueen/auth/common.py
@@ -13,10 +13,10 @@ import logging
 logger = logging.getLogger('kqueen_api')
 
 
-def generate_auth_config(auth_list):
+def generate_auth_options(auth_list):
     auth_options = {}
 
-    methods = auth_list.split(',')
+    methods = auth_list.split(',').strip()
     modules = AuthModules()
     for m in methods:
         if hasattr(modules, m):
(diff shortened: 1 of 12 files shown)


kqueen/config/auth.py
```diff
@@ -0,0 +1,25 @@
+from kqueen.config import current_config
+
+
+class AuthModules():
```
atengler: Is there a reason to create this class?

kqueen/config/utils.py
```diff
@@ -37,11 +36,14 @@ def apply_env_changes(config, prefix='KQUEEN_'):
         config (obj): Connfiguration object. This configuration will updated.
         prefix (str): Prefix for environment variables. Defaults to `KQUEEN_`.
 
+    return config_file
     """
 
     for name, value in os.environ.items():
         if name.startswith(prefix):
             config_key_name = name[len(prefix):]
+            if re.search('(?i)true|(?i)false', value):
```
atengler: If we are converting strings to bool, can't we convert json string to list/dict in python and avoid creating unnecessary class?

naumvd95: no, defining huge Json lists as Environment Variable for docker - is a bad case
variables should be simple and clear, all dict* stuff should be stored in configs

kqueen/config/auth.py
naumvd95: to increase visibility and avoid case of multiple value-entries in config file, when 
a = 'One'
b = a + 'Two' == 'One' + 'Two'

apply_changes:
a == 'EnvOne'
b == 'One' + 'Two'

kqueen/auth/common.py
```diff
@@ -12,18 +13,36 @@
 logger = logging.getLogger('kqueen_api')
 
 
+def generate_auth_config(auth_list):
+    auth_options = {}
+
+    methods = auth_list.split(',')
```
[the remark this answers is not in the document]
naumvd95: good point, thanks!

kqueen/auth/common.py
```diff
@@ -12,18 +13,36 @@
 logger = logging.getLogger('kqueen_api')
 
 
+def generate_auth_config(auth_list):
```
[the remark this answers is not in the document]
naumvd95: okok

kqueen/config/auth.py
naumvd95: we already discussed the case with u previously

naumvd95: and about class, i wanted to provide simple possibility to add another auth modules, if u have any other good solutions, please go ahead

kqueen/auth/common.py
```diff
@@ -12,18 +13,36 @@
 logger = logging.getLogger('kqueen_api')
```
katyafervent: you can set  
AUTH_MAP =
 {'ldap' = {"engine": "LDAPAuth",
+            "param": {
+                "uri": config.get('LDAP_URI')
+            }
+            }
+    'local' = {"engine": "LocalAuth",
+             "param": {}
+             }
}

here and not to create a new class

PR #263, opened 2018-04-05
Add 'order' to the engine UI parameter schema

katyafervent:
It will be useful to make form filling user friendly.
Thus, non-required field can be placed to the end of the form.


atengler:
Is it intentional that you do not add `order` for GCE engine?

katyafervent:
It's just one parameter there, but if you advice me to do that, I'll do

atengler:
Yes, my bad, its not necessary for GCE with defaulting to 0 in the UI PR

PR #264, opened 2018-04-05
Kqueen ldap

naumvd95:
refactor env-vars arch to provide configuration using only docker env-vars
remove apply_env changes due issues with bad overriding in case of multiple value-entries in config file
logging increased
splitted on separate commits to increase visibility


kqueen/auth/common.py
```diff
@@ -9,21 +9,66 @@
...
+    }
+}
+
+
+def generate_auth_options(auth_list):
+    auth_options = {}
+
+    methods = auth_list.strip().split(',')
+    for m in methods:
+        if AUTH_MODULES[m]:
```
atengler: `methods` is list of strings so you should do `if m in AUTH_MODULES`, this condition would actually fail with `KeyError` if `m` is not inside `AUTH_MODULES`

kqueen/auth/common.py
```diff
@@ -9,21 +9,66 @@
...
+    "ldap": {
+        "engine": "LDAPAuth",
+        "param": {
+            "uri": config.get('LDAP_URI')
+        }
+    },
+    "local": {
+        "engine": "LocalAuth",
+        "param": {
+
```
atengler: This line shouldn't be blank I think, check PEP8 pls

kqueen/auth/common.py
```diff
@@ -9,21 +9,66 @@
...
+                           "param": {
+                               "key": "value"
+                           }
+                           }
+"""
+
+AUTH_MODULES = {
+    "ldap": {
+        "engine": "LDAPAuth",
+        "param": {
```
atengler: Maybe use `params` instead of `param` or just full `parameters`. It seems that it was this way before this PR, but while you are refactoring, you could change this.

kqueen/auth/common.py
```diff
@@ -9,21 +9,66 @@
...
 
+config = current_config()
 logger = logging.getLogger('kqueen_api')
 
+"""
+Authentication Modules
+
+To define new module, need to specify it as dictionary, where:
+
+"auth_option_lower_case": {"engine": "EqualsToAuthClassName",
```
atengler: Just a detail, but fix the indentation pls

kqueen/auth/common.py
```diff
@@ -9,21 +9,66 @@
...
+    logger.debug('Auth config generated {}'.format(auth_options))
+    return auth_options
+
 
 def get_auth_instance(name):
     # Default type is local auth
 
     config = current_config()
-    auth_config = config.get("AUTH", {}).get(name, {})
+    auth_config = generate_auth_options(config.get("AUTH_MODULES")).get(name, {})
```
atengler: Maybe split this into two lines just to improve the readability a bit:
```
    auth_options = generate_auth_options(config.get('AUTH_MODULES'))
    auth_config = auth_options.get(name, {})
```

kqueen/config/utils.py
```diff
@@ -42,6 +43,8 @@ def apply_env_changes(config, prefix='KQUEEN_'):
     for name, value in os.environ.items():
         if name.startswith(prefix):
             config_key_name = name[len(prefix):]
+            if re.search('(?i)true|(?i)false', value):
```
atengler: I still think that the rest of this PR could be entirely avoided if you tested somehow that the env variable is supposed to be list/dict and used `json.load` on the next line here. But it is just a suggestion, I don't have any other solution, so I will approve if you just fix the stuff from other comments.

kqueen/auth/common.py
naumvd95: good point thanks!

kqueen/auth/common.py
naumvd95: flake passed, let me check

kqueen/auth/common.py
naumvd95: done

kqueen/auth/common.py
naumvd95: done

kqueen/auth/common.py
naumvd95: okko

kqueen/config/utils.py
naumvd95: it supposed, but IMO and for my experience, there is really bad case to define dicts in env-vars, coz there a lot of dependencies, for example if u compile docker stack arch, using reclass+salt+yml, smth can went wrong with dict syntax, so i prefer do not use that, but we can file a us and investigate it in future

kqueen/auth/common.py
```diff
@@ -9,28 +9,71 @@
...
 import logging
 
+config = current_config()
 logger = logging.getLogger('kqueen_api')
 
+"""
+    Authentication Modules
+
+    To define new module, need to specify it as dictionary, where:
+    "auth_option_lower_case": {"engine": "EqualsToAuthClassName",
```
atengler: I would still like this dict to be formatted properly, it just looks bad, sry :)

naumvd95: ah, let me change)

PR #265, opened 2018-04-06
Remove useless config defining method

naumvd95:
- To avoid multiple config calls and redundant logging
- To provide visibility of using configuration in docker container
- To avoid issues with overriding defined config in backend


naumvd95:
there a lot of testing dependencies, so u can just review it for now, and i prefer turn it into backlog

PR #266, opened 2018-04-09
Refactor Ldap Search-verify mechanism

naumvd95:
Refactor Ldap Search-verify mechanism

    - Search by provided dn segments to get full dn
    - possibility of adding full dn to user metadata
    - logging
    - docs updated
    - env variables for read-only user for Ldap-search

    Ldap-user adding scheme will be:

    UI:
    invite ldap user with cn 'kqueen'

    API:
    1. get dc's from predefined kqueen-ldap user
    2. search in all OU by cn and dc's and paste all matched cases into list

    UI:
    try to enter login and text 'password'

    API:
    3. try to bind_s with list of mathced_dn and provided 'password'
    4. in case of success, store valid full-dn into user-metadata, and re-use it without search-duplicating

    P.s. afaik in LDAP user 'CN' should be absolutely equal.Its absolutely independent of OU,Posix groups.
    Ldap doesnot allow to create users in diff groups with equal 'CN'. So , it's avoid us of several searching problems
PAY ATTENTION:
This PR depends on UI changes related to Invite form improvement @katyafervent

### DIFF
--- a/kqueen/auth/common.py
+++ b/kqueen/auth/common.py
@@ -28,7 +28,9 @@ AUTH_MODULES = {
     "ldap": {
         "engine": "LDAPAuth",
         "parameters": {
-            "uri": config.get('LDAP_URI')
+            "uri": config.get('LDAP_URI'),
+            "admin_dn": config.get('LDAP_DN'),
+            "password": config.get('LDAP_PASSWORD')
         }
     },
     "local": {
@@ -96,10 +98,9 @@ def authenticate(username, password):
     user = username_table.get(username)
 
     if user:
+        user.metadata = user.metadata or {}
         given_password = password.encode('utf-8')
-
         logger.debug("User {} will be authenticated using {}".format(username, user.auth))
-
         auth_instance = get_auth_instance(user.auth)
 
         try:
--- a/kqueen/auth/ldap.py
+++ b/kqueen/auth/ldap.py
@@ -1,6 +1,7 @@
 from .base import BaseAuth
 
 from kqueen.config import current_config
+from kqueen.exceptions import ImproperlyConfigured
 import ldap
 import logging
 
@@ -15,38 +16,46 @@ class LDAPAuth(BaseAuth):
         """
 
         super(LDAPAuth, self).__init__(*args, **kwargs)
+        if not all(hasattr(self, attr) for attr in ['uri', 'admin_dn', 'password']):
+            msg = 'Failed to configure LDAP, please provide valid LDAP credentials'
+            logger.error(msg)
+            raise ImproperlyConfigured(msg)
 
-        if not hasattr(self, 'uri'):
-            raise Exception('Parameter uri is required')
+        # Define Kqueen rdn for all dc's
+        d_names = ldap.dn.explode_dn(self.admin_dn)
+        dc_list = [dc for dc in d_names if dc.startswith('dc=')]
+        self.kqueen_dc = ','.join(dc_list)
 
-        self.connection = ldap.initialize(self.uri)
+        # Bind connection for Kqueen Read-only user
+        if self._bind(self.admin_dn, self.password):
+            self.connection = ldap.initialize(self.uri)
+            self.connection.simple_bind_s(self.admin_dn, self.password)
+            self.connection.protocol_version = ldap.VERSION3
+        else:
+            logger.error('Failed to bind connection for Kqueen Read-only user')
+            self.connection.unbind()
 
-    @staticmethod
-    def _email_to_dn(email):
-        """This function reads email and converts it to LDAP dn
+    def _get_matched_dn(self, cn):
+        """This function reads username as cn and returns all matched full-dn's
 
         Args:
-            email (str): e-mail address
+            cn (str): Username of invited user
 
         Returns:
-            dn (str): LDAP dn, like 'cn=admin,dc=example,dc=org'
+            matched_dn (list): List of all matched dn's in groups.
         """
 
-        segments = []
-        if '@' in email:
-            cn, dcs = email.split('@')
-        else:
-            cn = email
-            dcs = ''
+        base_dn = self.kqueen_dc
+        search_scope = ldap.SCOPE_SUBTREE
+        retrieveAttributes = ['dn']
+        search_filter = "(cn={})".format(cn)
+        search_result = self.connection.search_s(base_dn, search_scope, search_filter, retrieveAttributes)
 
-        if cn:
-            segments.append('cn={}'.format(cn))
+        matched_dn = [dn[0] for dn in search_result]
+        logger.info('Matched to RegExp DN key: {}'.format(matched_dn))
 
-        if '.' in dcs:
-            for s in dcs.split('.'):
-                segments.append('dc={}'.format(s))
-
-        return ','.join(segments)
+        self.connection.unbind()
+        return matched_dn
 
     def verify(self, user, password):
         """Implementation of :func:`~kqueen.auth.base.__init__`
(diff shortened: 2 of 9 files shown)


kqueen/auth/ldap.py
```diff
@@ -22,31 +22,62 @@ def __init__(self, *args, **kwargs):
...
+        segments['dcs'] = ','.join(dc_list)
+        logger.debug('RegExp for search valid full-DN for Ldap user: {}'.format(segments))
+
+        return segments
+
+    def _get_matched_dn(self, username):
+        dn_segments = self._email_to_dn_dict(username)
+
+        if self._bind(self.kqueen_username, self.kqueen_password):
+            self.connection = ldap.initialize(self.uri)
```
katyafervent: it should be already evaluated in init method, so you can just use self.connection right away

kqueen/auth/ldap.py
```diff
@@ -22,31 +22,62 @@ def __init__(self, *args, **kwargs):
...
+            self.connection.simple_bind_s(self.kqueen_username, self.kqueen_password)
+            self.connection.protocol_version = ldap.VERSION3
```
katyafervent: it should be set in the init method with the comment why it is needed

kqueen/auth/ldap.py
```diff
@@ -22,31 +22,62 @@ def __init__(self, *args, **kwargs):
...
+
+            baseDN = dn_segments['dcs']
```
katyafervent: it is not a good name for a variable regarding Python convention

kqueen/auth/ldap.py
```diff
@@ -22,31 +22,62 @@ def __init__(self, *args, **kwargs):
...
+            searchScope = ldap.SCOPE_SUBTREE
```
katyafervent: same as above

kqueen/auth/ldap.py
```diff
@@ -22,31 +22,62 @@ def __init__(self, *args, **kwargs):
...
+            matched_dn = []
+            for key, value in dict(ldap_result_id).items():
+                matched_dn.append(key)
+
+            logger.debug('All matched to RegExp DN keys: {}'.format(matched_dn))
+            self.connection.unbind()
+            return matched_dn
+
+        else:
+            logger.debug('Failed to get matched to RegExp {} DN due connection problems'.format(matched_dn))
```
katyafervent: looks like error, I guess we need to raise an exception here or at list set error log level. Is it handled above?

kqueen/auth/ldap.py
```diff
@@ -61,30 +92,42 @@ def verify(self, user, password):
...
+            logger.debug('Valid full-DN found: {}. It will be stored in user metadata'.format(full_dn))
+            return user, None
+
+        return None, "Verification failed"
 
-        dn = self._email_to_dn(user.username)
+    def _bind(self, dn, password):
 
         try:
+            self.connection = ldap.initialize(self.uri)
```
katyafervent: already initialized in the constructor

kqueen/auth/ldap.py
```diff
@@ -61,30 +92,42 @@ def verify(self, user, password):
...
             bind = self.connection.simple_bind_s(dn, password)
 
             if bind:
-                return user, None
+                return True
```
katyafervent: Initially this method return username and error message, you reworked it and now it is returned two or one value, which is inappropriate

kqueen/auth/ldap.py
```diff
@@ -61,30 +92,42 @@ def verify(self, user, password):
...
             logger.exception("Invalid DN syntax in configuration: {}".format(dn))
 
-            return None, "Invalid DN syntax"
+            return False, "Invalid DN syntax"
 
         except ldap.LDAPError:
             logger.exception("Failed to bind LDAP server")
 
-            return None, "LDAP auth failed, check log for error"
+            return False, "LDAP auth failed, check log for error"
```
atengler: You might want to handle generic `Exception` in the end, because you might lose network connection for example

kqueen/auth/ldap.py
```diff
@@ -61,30 +92,42 @@ def verify(self, user, password):
             None: Invalid credentials, Authentication failure.
             User: authenticated user.
         """
+        matched_dn = self._get_matched_dn(user.username)
+
+        for dn in matched_dn:
+            if self._bind(dn, password):
+                full_dn = dn
+
+        if full_dn:
```
atengler: You are defining `full_dn` variable inside `if` condition so it may not be defined if none of the matched users authenticate with given password, it would raise `referenced before assignment` error. You should prolly add `full_dn = None` above the `for` loop. Also I don't see you actually storing the verified `full_dn` on User object and checking for it in the beginning of `verify` method to skip the search.

kqueen/auth/ldap.py
```diff
@@ -22,31 +22,62 @@ def __init__(self, *args, **kwargs):
...
-        """This function reads email and converts it to LDAP dn
+    def _email_to_dn_dict(email):
+        """This function reads email and converts it to LDAP dn and dc's
 
         Args:
             email (str): e-mail address
 
         Returns:
-            dn (str): LDAP dn, like 'cn=admin,dc=example,dc=org'
+            dn (dict): LDAP dn and dc's, like {'cn': 'cn=admin', 'dcs': 'dc=example,dc=org'}
```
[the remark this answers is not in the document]
naumvd95: hm, your point is not clear for me, pls clarify info, i need to change Return value naming?

kqueen/auth/ldap.py
naumvd95: no, previously we use unbind() only once, and all works fine
but in this case we use unbind multiple times, and each time we should re-init ldap object, otherwise if=t failed to bind. i can clarify info directly or take a look on LDAP docs

kqueen/auth/ldap.py
naumvd95: good point, sorry, it was copied from basic tutors, ill switch it to snake_case

kqueen/auth/ldap.py
naumvd95: good point, sorry, it was copied from basic tutors, ill switch it to snake_case

kqueen/auth/ldap.py
naumvd95: it will handled by _bind() method exceptions

kqueen/auth/ldap.py
naumvd95: no, previously we use unbind() only once, and all works fine
but in this case we use unbind multiple times, and each time we should re-init ldap object, otherwise if=t failed to bind. i can clarify info directly or take a look on LDAP docs

kqueen/auth/ldap.py
naumvd95: oh, thanks, i need to switch all return to 1-value arch, yes?

kqueen/auth/ldap.py
naumvd95: good point, thanks! about metadata storing, before my changes, i wanted to get your review and discuss, so it does not exist yet

kqueen/auth/ldap.py
naumvd95: seems your point is not clear for me , pls clarify info

kqueen/auth/ldap.py
naumvd95: good point, thx

kqueen/auth/ldap.py
atengler: You handle various LDAP errors in the try/except block, but if you would for example lose network connection and you are unable to reach the LDAP server, then try block fails, but none of these except blocks will catch the `unreachable` exception. So I would add `except Exception as e` in the end and create message like `msg = 'Unknown error occurred: {}'.format(repr(e))` or something. So I handle even the exceptions you don't necessarily expect to occur

kqueen/auth/common.py
```diff
@@ -28,7 +28,9 @@
     "ldap": {
         "engine": "LDAPAuth",
         "parameters": {
-            "uri": config.get('LDAP_URI')
+            "uri": config.get('LDAP_URI'),
+            "kqueen_username": config.get('LDAP_USERNAME'),
```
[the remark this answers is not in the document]
katyafervent: just username or better DN, you don't hane to create special user for kqueen.

kqueen/auth/ldap.py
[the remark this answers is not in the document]
katyafervent: https://www.python-ldap.org/en/latest/reference/ldap-dn.html#ldap.dn.str2dn

yes, check return value - we can return list of tuples or dictionary of lists

kqueen/auth/ldap.py
naumvd95: done

kqueen/auth/common.py
```diff
@@ -28,7 +28,9 @@
     "ldap": {
         "engine": "LDAPAuth",
         "parameters": {
-            "uri": config.get('LDAP_URI')
+            "uri": config.get('LDAP_URI'),
+            "kqueen_dn": config.get('LDAP_DN'),
+            "kqueen_password": config.get('LDAP_PASSWORD')
```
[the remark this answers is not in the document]
naumvd95: ok

kqueen/auth/ldap.py
```diff
@@ -16,37 +17,49 @@ def __init__(self, *args, **kwargs):
 
         super(LDAPAuth, self).__init__(*args, **kwargs)
 
-        if not hasattr(self, 'uri'):
-            raise Exception('Parameter uri is required')
-
-        self.connection = ldap.initialize(self.uri)
+        if not all(hasattr(self, attr) for attr in ['uri', 'kqueen_dn', 'kqueen_password']):
+            raise ImproperlyConfigured
```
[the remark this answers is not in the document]
naumvd95: IMO exception class refactoring is not depended to this PR, so i can just log this

kqueen/auth/ldap.py
```diff
@@ -15,38 +16,50 @@ def __init__(self, *args, **kwargs):
         """
 
         super(LDAPAuth, self).__init__(*args, **kwargs)
+        if not all(hasattr(self, attr) for attr in ['uri', 'admin_dn', 'password']):
+            logger.error('Failed to configure LDAP synchronization error.')
+            raise ImproperlyConfigured
+
+        # Define Kqueen rdn for all dc's
+        dc_list = []
```
atengler: Maybe replace this with something like:
```
d_names = ldap.dn.explode_dn(self.admin_dn)
dc_list = [dc for dc in d_names if dc.startswith('dc=')
```
Feels a bit better and I don't really like using single letter variable names outside of generator like this

kqueen/auth/ldap.py
```diff
@@ -15,38 +16,50 @@ def __init__(self, *args, **kwargs):
...
+        base_dn = self.kqueen_dc
+        search_scope = ldap.SCOPE_SUBTREE
+        retrieveAttributes = ['dn']
+        search_filter = "(cn={})".format(cn)
+        search_result = self.connection.search_s(base_dn, search_scope, search_filter, retrieveAttributes)
 
-        if '.' in dcs:
-            for s in dcs.split('.'):
-                segments.append('dc={}'.format(s))
+        if search_result:
```
atengler: You don't need to test if `search_result` is empty, the generator will simply end up being empty list if it is. If you do it this way, you are again facing the issue, that `matched_dn` may not be defined when you want to return it, so it will raise error

kqueen/auth/test_ldap.py
```diff
@@ -8,18 +9,20 @@ class TestAuthMethod:
     @pytest.fixture(autouse=True)
     def setup(self, user):
         self.user = user
-        self.user.username = 'admin@example.org'
+        self.user.username = 'admin'
+        self.user.metadata = {}
```
katyafervent: Should we add test with fulled metadata?

PR #267, opened 2018-04-11
Add configurations/auth API call

katyafervent:
So it will be possible to know witch authitenfication methods are used

*todo: once we add ldap admin and password for ldap thoese paramets should be hidden


kqueen/auth/common.py
```diff
@@ -45,9 +45,10 @@ def generate_auth_options(auth_list):
     for m in methods:
         if m in AUTH_MODULES:
             auth_options[m] = AUTH_MODULES[m]
-
+        else:
+            logger.debug('Specified in config auth method {} is not found in the auth modules specification ')
```
naumvd95: .format() is missed

kqueen/auth/common.py
```diff
@@ -45,9 +45,10 @@ def generate_auth_options(auth_list):
...
     if not auth_options:
```
naumvd95: seems need to refactor this, coz we already call ```             auth_options[m] = AUTH_MODULES[m]
``` on raw 47

File: kqueen/engines/base.py [hunk 3]
```diff
@@ -187,6 +187,9 @@ def get_parameter_schema(cls):
         Returns:
             dict:  Returns ``self.parameter_schema`` in default, but can be overridden.
         """
+        if not hasattr(cls, 'parameter_schema'):
+            raise NotImplementedError('"parameter_schema" attribute should be provided in the '
```
naumvd95: add logger before raising, coz this error does not traced on log by default
try to do:
```
msg = xxx
logger.debug(msg)
raise exception(msg)
```

naumvd95:
@katyafervent  pls change commit name ```
Make some improvements ```

'do smth/a little bit/ some features' its a bad practise to commit naming and searching by history
add some useful keywords

naumvd95:
i prefer u to check this PR with my ldap-search case, can u do that?

katyafervent:
What do you mean by check? We just need to perform merging later

kqueen/engines/base.py
katyafervent: it is logged where we catch it, i've already added logger there

kqueen/auth/common.py
katyafervent: good point

naumvd95:
@katyafervent does it tested?

PR #269, opened 2018-04-13
Calico networking support for GKE

naumvd95:
use-cases:
    1) User deploy GKE cluster and enable CALICO network policy:
    - network addon enabled too , due dependencies
    - network policy enabled
    2) User wants to on/off Network policy on existing cluster:
    - cluster must be deployed like p.1 only, otherwise, user cannot manage policies
    =============
    network-range
    use-cases:
    1) User deploy GKE cluster and specify network range:
    -The IP address range of the container pods in this cluster, in CIDR notation (e.g. 10.96.0.0/14).
    Leave blank to have one automatically chosen or specify a /14 block in 10.0.0.0/8.

    P.s.
    - refactoring to associate 100-symbols raw
    - logging increased

### DIFF
--- a/kqueen/blueprints/api/views.py
+++ b/kqueen/blueprints/api/views.py
@@ -65,7 +65,7 @@ def not_implemented(error):
 @api.route('/')
 @api.route('/health')
 def index():
-    return jsonify({'response': 'Gutten tag!'})
+    return jsonify({'response': 'Kqueen ready!'})
 
 
 # Clusters
--- a/kqueen/blueprints/api/views.py
+++ b/kqueen/blueprints/api/views.py
@@ -221,6 +221,30 @@ def cluster_resize(pk):
     return jsonify(output)
 
 
+@api.route('/clusters/<uuid:pk>/set_network_policy', methods=['PATCH'])
+@jwt_required()
+def cluster_set_network_policy(pk):
+    obj = get_object(Cluster, pk, current_identity)
+
+    if not request.json:
+        abort(400, description='JSON data expected')
+
+    data = request.json
+    if not all(k in data for k in ('provider', 'enabled')):
+        msg = 'Failed to get network policy configuration'
+        logger.error(msg)
+        abort(400, description=msg)
+
+    res_status, res_msg = obj.engine.set_network_policy(data['provider'], data['enabled'])
+    if not res_status:
+        logger.error('Setting network policy failed: {}'.format(res_msg))
+        abort(500, description=res_msg)
+
+    # get object with updated metadata
+    output = obj.engine.cluster
+    return jsonify(output)
+
+
 # Provisioners
 class ListProvisioners(ListView):
     object_class = Provisioner
--- a/kqueen/blueprints/api/views.py
+++ b/kqueen/blueprints/api/views.py
@@ -231,7 +231,7 @@ def cluster_set_network_policy(pk):
 
     data = request.json
     if not all(k in data for k in ('provider', 'enabled')):
-        msg = 'Failed to get network policy configuration'
+        msg = 'Incorrect network policy configuration {}'.format(data)
         logger.error(msg)
         abort(400, description=msg)
 
--- a/kqueen/config/base.py
+++ b/kqueen/config/base.py
@@ -29,7 +29,7 @@ class BaseConfig:
     CLUSTER_OK_STATE = 'OK'
     CLUSTER_PROVISIONING_STATE = 'Deploying'
     CLUSTER_DEPROVISIONING_STATE = 'Destroying'
-    CLUSTER_RESIZING_STATE = 'Resizing'
+    CLUSTER_UPDATING_STATE = 'Updating'
     CLUSTER_UNKNOWN_STATE = 'Unknown'
 
     CLUSTER_STATE_ON_LIST = True
--- a/kqueen/engines/base.py
+++ b/kqueen/engines/base.py
@@ -164,6 +164,21 @@ class BaseEngine:
         """
         raise NotImplementedError
 
+    def set_network_policy(self, network_provider, enabled):
+        """Set specific network policy to the cluster related to this engine instance.
+
+        Args:
+            network_provider(str):           Name of supported network provider/addon
+            enabled(bool):                   Enable/Disable policy
+
+        Returns:
+            tuple: First item is bool (success/failure), second item is error, can be None::
+
+                (True, None)                            # successful policy update
+                (False, 'error_message')                # failed policy update, error description
+        """
+        raise NotImplementedError
+
     def get_kubeconfig(self):
         """Get kubeconfig of the cluster related to this engine from backend.
 
--- a/kqueen/engines/gce.py
+++ b/kqueen/engines/gce.py
@@ -106,7 +106,7 @@ class GceEngine(BaseEngine):
             'network_range': {
                 'type': 'text',
                 'label': 'Network range CIDR',
-                'order': 5,
+                'order': 4,
                 'placeholder': '10.0.0.0/14',
                 'validators': {
                     'required': False,
@@ -125,7 +125,7 @@ class GceEngine(BaseEngine):
                 ],
                 'default': 'PROVIDER_UNSPECIFIED',
                 'validators': {
-                    'required': True
+                    'required': False
                 },
                 'class_name': 'network-policy'
             }
@@ -320,10 +320,10 @@ class GceEngine(BaseEngine):
 
         logger.debug('Setting {} network policy to cluster {}...'.format(network_provider,
                                                                          self.cluster_id))
-        request = self.client.projects().zones().clusters().setNetworkPolicy(projectId=self.project,
-                                                                             zone=self.zone,
-                                                                             clusterId=self.cluster_id,
-                                                                             body=network_policy_body)
+        request = self.client.projects().zones().clusters().setNetworkPolicy(
+            projectId=self.project, zone=self.zone,
+            clusterId=self.cluster_id, body=network_policy_body)
+
         try:
             request.execute()
         except Exception as e:
--- a/kqueen/engines/gce.py
+++ b/kqueen/engines/gce.py
@@ -35,6 +35,7 @@ class GceEngine(BaseEngine):
             'service_account_info': {
                 'type': 'json_file',
                 'label': 'Service Account File (JSON)',
+                'order': 0,
                 'validators': {
                     'required': True,
                     'jsonfile': [
@@ -52,7 +53,9 @@ class GceEngine(BaseEngine):
             'node_count': {
                 'type': 'integer',
                 'label': 'Node Count',
+                'order': 1,
                 'default': 1,
+                'class_name': 'gke_node_count',
                 'validators': {
                     'required': True,
                     'min': 1,
@@ -62,6 +65,7 @@ class GceEngine(BaseEngine):
             'zone': {
                 'type': 'select',
                 'label': 'Zone',
+                'order': 2,
                 'choices': [
                     ('us-central1-a', 'US - Central 1 - A'),
                     ('us-west1-a', 'US - West 1 - A'),
@@ -85,6 +89,7 @@ class GceEngine(BaseEngine):
             'machine_type': {
                 'type': 'select',
                 'label': 'Machine Type',
+                'order': 3,
                 'choices': [
                     ('n1-standard-1', 'Standart: 1 vCPU, 3.75 GB RAM'),
                     ('n1-standard-2', 'Standart: 2 vCPU, 7.5 GB RAM'),
@@ -97,6 +102,29 @@ class GceEngine(BaseEngine):
                 'validators': {
                     'required': True
                 }
+            },
+            'network_range': {
+                # TODO set CIDR validators, decide class naming after UI changes
+                'type': 'text',
+                'label': 'Network range CIDR',
+                'order': 4,
+                'class_name': 'gke_network_range',
+                'validators': {
+                    'required': False,
+                }
+            },
+            'network_policy': {
+                'type': 'select',
+                'label': 'Network Policy',
+                'order': 5,
+                'choices': [
+                    (None, '(None)'),
+                    ('CALICO', 'Calico')
+                ],
+                'validators': {
+                    'required': True
+                },
+                'class_name': 'network-policy'
             }
         }
     }
@@ -118,9 +146,22 @@ class GceEngine(BaseEngine):
                 'initialNodeCount': kwargs.get('node_count', 1),
                 'nodeConfig': {
                     'machineType': kwargs.get('machine_type', 'n1-standard-1')
+                },
+                'addonsConfig': {},
+                'clusterIpv4Cidr': kwargs.get('network_range', ''),
+                'networkPolicy': {
+                    'provider': kwargs['network_policy'].get('provider', 'PROVIDER_UNSPECIFIED'),
+                    'enabled': bool(kwargs.get('network_policy', False))
                 }
             }
         }
+        if self.cluster_config['cluster']['networkPolicy']['enabled'] is True:
+            logger.debug('Network addon for GKE enabled')
+            self.cluster_config = self._set_addon_config(cluster_config=self.cluster_config,
+                                                         addon='networkPolicyConfig',
+                                                         disabled=False)
+
+        logger.debug('GKE cluster configuration: {}'.format(self.cluster_config))
         self.client = self._get_client()
         # Cache settings
         self.cache_timeout = 5 * 60
@@ -136,17 +177,55 @@ class GceEngine(BaseEngine):
 
         return client
 
+    def _set_addon_config(self, cluster_config, addon, disabled):
+        """Set addon configutation to the cluster.
+
+        Args:
+            cluster_config(dict): Current cluster configuration
+            addon(str):           Name of supported addon
+            disabled(bool):       Enable/Disable addon
+
+        Returns:
+            dict:                 Updated cluster configuration
+
+        """
+        addons_body = {
+            addon: {
+                'disabled': disabled
+            }
+        }
+        addons_config = cluster_config['cluster'].get('addonsConfig', {})
+        addons_config[addon] = addons_body[addon]
+        logger.debug('Setting {} addon in cluster_config {}'.format(addon, cluster_config))
+
+        return cluster_config
+
     def provision(self, **kwargs):
         """
         Implementation of :func:`~kqueen.engines.base.BaseEngine.provision`
         """
+
+        request = self.client.projects().zones().clusters().create(projectId=self.project,
+                                                                   zone=self.zone,
+                                                                   body=self.cluster_config)
         try:
-            self.client.projects().zones().clusters().create(projectId=self.project, zone=self.zone, body=self.cluster_config).execute()
+            request.execute()
             # TODO: check if provisioning response is healthy
         except Exception as e:
-            msg = 'Creating cluster {} failed with following reason:'.format(self.cluster_id)
+            msg = 'Creating cluster {} failed with the following reason: {}'.format(self.cluster_id,
+                                                                                    e)
             logger.exception(msg)
             return False, msg
+
+        current_policy = self.cluster_config['cluster']['networkPolicy']
+        meta = self.cluster.metadata.get('network_policy', {})
+
+        if current_policy['provider'] is not None:
+            meta['provider'] = current_policy['provider']
+            meta['enabled'] = current_policy['enabled']
+            logger.critical('Provisioning cluster {} started,\
+                            updating metadata...{}'.format(self.cluster_id, meta))
+            self.cluster.save()
 
         return True, None
 
@@ -158,18 +237,31 @@ class GceEngine(BaseEngine):
         result, error = super(GceEngine, self).deprovision(**kwargs)
         if result:
             return result, error
-
+        request = self.client.projects().zones().clusters().delete(projectId=self.project,
+                                                                   zone=self.zone,
+                                                                   clusterId=self.cluster_id)
         try:
-            self.client.projects().zones().clusters().delete(projectId=self.project, zone=self.zone, clusterId=self.cluster_id).execute()
+            request.execute()
             # TODO: check if provisioning response is healthy
         except Exception as e:
-            msg = 'Deleting cluster {} failed with following reason: {}'.format(self.cluster_id, repr(e))
+            msg = 'Deleting cluster {} failed with following reason: {}'.format(self.cluster_id,
+                                                                                repr(e))
             logger.exception(msg)
             return False, msg
 
         return True, None
 
     def resize(self, node_count, **kwargs):
+
+        if int(node_count) < 2 and \
+           self.cluster_config['cluster']['networkPolicy']['enabled'] is True:
+            msg = 'Resizing cluster {} denied. The minimum size cluster to run \
+                   network policy enforcement is 2 n1-standard-1 instances.\
+                   Otherwise, turn off network policy before resizing.'\
+                   .format(self.cluster_id)
+            logger.error(msg)
+            return False, msg
+
         request = self.client.projects().zones().clusters().nodePools().setSize(
             nodePoolId='default-pool',
             clusterId=self.cluster_id,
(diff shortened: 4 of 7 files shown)


kqueen/engines/gce.py
```diff
@@ -189,6 +200,34 @@ def resize(self, node_count, **kwargs):
...
+        }
+
+        if self.cluster.metadata['node_count'] < 2 or self.cluster_config['cluster']['nodeConfig']['machineType'] in unsupported_instances:
+            msg = 'Setting {} Network Policy for cluster {} denied due unsupported configuration. \
+                   The recommended minimum size cluster to run network policy enforcement is 3 n1-standard-1 instances'.format(network_provider, self.cluster_id)
+            logger.error(msg)
+            return False, msg
+
+        try:
+            self.client.projects().zones().clusters().setNetworkPolicy(projectId=self.project, zone=self.zone, clusterId=self.cluster_id, body=network_policy_body).execute()
```
ekhomyakova: Could you not make such a long lines. For python, max length is 79, but if you want more, at least if should fit the window

naumvd95: good point, thx

katyafervent: is this json? you have to add json generation function, since in ui would be only text fields or whatever is required

kqueen/blueprints/api/views.py
```diff
@@ -220,6 +220,27 @@ def cluster_resize(pk):
...
 
 
+@api.route('/clusters/<uuid:pk>/set_network_policy', methods=['PATCH'])
+@jwt_required()
+def cluster_set_network_policy(pk):
+    obj = get_object(Cluster, pk, current_identity)
+
+    data = request.json
+    if not isinstance(data, dict) or (isinstance(data, dict) and 'node_count' not in data):
+        abort(400)
```
katyafervent: please, provide a description to the abort function

kqueen/blueprints/api/views.py
```diff
@@ -220,6 +220,27 @@ def cluster_resize(pk):
...
+
+    res_status, res_msg = obj.engine.set_network_policy(data['network_policy']['provider'],
```
katyafervent: what is 'network_policy' not in data?

kqueen/engines/gce.py
```diff
@@ -189,6 +204,43 @@ def resize(self, node_count, **kwargs):
 
         return True, None
 
+    def set_network_policy(self, network_provider='CALICO', enabled=False):
```
katyafervent: please, add a docstring with parameter and return values description and why it is disabled by default

kqueen/engines/gce.py
```diff
@@ -189,6 +204,43 @@ def resize(self, node_count, **kwargs):
...
+        network_policy_body = {
+            'networkPolicy': {
+                'provider': network_provider,
+                'enabled': enabled
+            }
+        }
+        m_type = self.cluster['nodeConfig']['machineType']
+        if self.cluster.metadata['node_count'] < 2 or \
+           m_type in unsupported_instances:
+            msg = 'Setting {} Network Policy for cluster {} denied due \
```
katyafervent: for cluster -> for the cluster
due -> due to

kqueen/engines/gce.py
```diff
@@ -170,6 +174,16 @@ def deprovision(self, **kwargs):
         return True, None
 
     def resize(self, node_count, **kwargs):
+
+        if node_count < 2 and \
+           self.cluster_config['cluster']['networkPolicy']['enabled'] is True:
```
katyafervent: cluster may not have 'networkPolicy' so you should check it with _get_ before trying to get enabled

kqueen/blueprints/api/views.py
```diff
@@ -220,6 +220,27 @@ def cluster_resize(pk):
...
+                                                        data['network_policy']['enabled'])
+
+    if not res_status:
+        logger.error('Setting network policy failed: {}'.format(res_msg))
+        abort(500, description=res_msg)
```
katyafervent: Actually, I don't think that 500 error is a good choice here. It indicates with problems on server side, but if this action is unappropriate due to he recommended minimum size \
+                   cluster to run network policy enforcement is 3 \
+                   n1-standard-1 instances or small node count, the respond code should be forbidden.

kqueen/blueprints/api/views.py
```diff
@@ -220,6 +220,27 @@ def cluster_resize(pk):
     return jsonify(output)
 
 
+@api.route('/clusters/<uuid:pk>/set_network_policy', methods=['PATCH'])
+@jwt_required()
+def cluster_set_network_policy(pk):
+    obj = get_object(Cluster, pk, current_identity)
+
+    data = request.json
+    if not isinstance(data, dict) or (isinstance(data, dict) and 'node_count' not in data):
```
ekhomyakova: > or (isinstance(data, dict) and 'node_count' not in data)
> isinstance(data, dict)
This is redundant, because you will not reach this condition if otherwise

kqueen/engines/gce.py
naumvd95: it should be anyway, coz i defined network policy in config on __init__ method

kqueen/blueprints/api/views.py
naumvd95: why not? pls clarify info

kqueen/blueprints/api/views.py
naumvd95: and what do u prefer?

kqueen/blueprints/api/views.py
ekhomyakova:
```
not x or (x and y) = not x or y

x | y | not x | x and y | not x or (x and y) | not x or y
0 | 0 |   1   |    0    |         1          |      1    
0 | 1 |   1   |    0    |         1          |      1    
1 | 0 |   0   |    0    |         0          |      0    
1 | 1 |   0   |    1    |         1          |      1    
```
Hope this is visually enough

ekhomyakova: Or `not x or (x and y) = not x or x and not x or y = True and not x or y = not x or y`

kqueen/engines/gce.py
```diff
@@ -118,6 +118,10 @@ def __init__(self, cluster, **kwargs):
                 'initialNodeCount': kwargs.get('node_count', 1),
                 'nodeConfig': {
                     'machineType': kwargs.get('machine_type', 'n1-standard-1')
+                },
+                'networkPolicy': {
+                    'provider': kwargs.get('network_provider', 'CALICO'),
+                    'enabled': kwargs.get('network_policy_enabled', True)
```
atengler: So, when you pass `True` here, Calico Policy Engine will be automatically deployed into the cluster (from my understanding its k8s pod/service running inside the cluster) after its created. If you pass `False` instead, it won't be, but you can add it later with the `set_network_policy` method, is that correct? If so, can you create `unset_network_policy` which removes the policy as well?

atengler:
I think that what you do is mostly correct, I left a question for clarification on the cluster metadata dict, but you should define all of the new methods added to GKE cluster also to base class, because in perfect world we should create generic functionality which can be applied on all(/most) provisioners. You can just create the methods with the same names and same args/kwargs under this class - https://github.com/Mirantis/kqueen/blob/master/kqueen/engines/base.py#L9 , document them in docstring under method definition and let them raise `NotImplementedError` or return False or whatever suits the particular method when not implemented.

kqueen/blueprints/api/views.py
```diff
@@ -220,6 +220,29 @@ def cluster_resize(pk):
     return jsonify(output)
 
 
+@api.route('/clusters/<uuid:pk>/set_network_policy', methods=['PATCH'])
+@jwt_required()
+def cluster_set_network_policy(pk):
+    obj = get_object(Cluster, pk, current_identity)
+
+    data = request.json
+    if not isinstance(data, dict) or (isinstance(data, dict) and 'network_policy' not in data):
```
ekhomyakova: Please take a look at my previous comments

kqueen/engines/gce.py
```diff
@@ -189,6 +204,55 @@ def resize(self, node_count, **kwargs):
 
         return True, None
 
+    def set_network_policy(self, network_provider='CALICO', enabled=False):
```
atengler: If I understand correctly what the method does, then you are not setting network policy in this method, you are enabling the possibility to set network policies, so I guess I would rename the method to `enable_network_policy`

kqueen/engines/gce.py
```diff
@@ -189,6 +204,55 @@ def resize(self, node_count, **kwargs):
...
+            network_provider (str): Name of supported network provider.
+            Currently GKE support only Calico driver.
+
+            enabled (bool): Enabled preferred policy, and deploy additional
+            k8s services, related to network managment.
+
+        Returns:
+             (bool): Result of policy-update
+        """
+
```
ekhomyakova: Please remove the empty line

kqueen/engines/gce.py
```diff
@@ -189,6 +204,55 @@ def resize(self, node_count, **kwargs):
...
+                'provider': network_provider,
+                'enabled': enabled
+            }
+        }
+
+        m_type = self.cluster['nodeConfig']['machineType']
+        if self.cluster.metadata['node_count'] < 2 or \
+           m_type in unsupported_instances:
+            msg = 'Setting {} Network Policy for the cluster {} denied due to \
+                   unsupported configuration. The recommended minimum size \
```
ekhomyakova: Maybe there should be also info about required size cluster

kqueen/engines/gce.py
```diff
@@ -189,6 +204,55 @@ def resize(self, node_count, **kwargs):
...
+
+        try:
+            self.client.projects().zones().clusters().\
+                setNetworkPolicy(projectId=self.project,
+                                 zone=self.zone,
+                                 clusterId=self.cluster_id,
+                                 body=network_policy_body).execute()
+        except Exception as e:
+            msg = 'Setting {} Network Policy for cluster {} failed with \
+                   following reason:'.format(network_provider, self.cluster_id)
```
ekhomyakova: the following

kqueen/engines/gce.py
```diff
@@ -180,7 +194,8 @@ def resize(self, node_count, **kwargs):
         try:
             request.execute()
         except Exception as e:
-            msg = 'Resizing cluster {} failed with following reason: {}'.format(self.cluster_id, repr(e))
+            msg = 'Resizing cluster {} failed with following reason: {}'\
```
ekhomyakova: the following

kqueen/blueprints/api/views.py
```diff
@@ -220,6 +220,29 @@ def cluster_resize(pk):
...
+        msg = 'Failed to get network policy configuration'
+        logger.error(msg)
+        abort(400, description=msg)
+
+    res_status, res_msg = obj.engine.set_network_policy(data['network_policy']['provider'],
```
ekhomyakova: Adding of `network_policy` key makes sense only if you pass it with cluster creation. Otherwise it is redundant as you send a request to `set_network_policy`

ekhomyakova: By the way, could you add an ability to set policy with cluster creation? It's easier when you do only one request

kqueen/blueprints/api/views.py
katyafervent: when user doesn't satisfy cluster to run network policy enforcement is 3  n1-standard-1 instances or 2 node count - 403 -forbidden, should be return, not 500. you can return it in set_network_policy as part of tuple that you are already returning

kqueen/engines/gce.py
katyafervent: So by default it will be turned on, are you sure?)

naumvd95: nono, its just draft) ill change it anyway to false by default, or refactor this dict-enabling method as @atengler said

kqueen/engines/gce.py
naumvd95: no, ideally this method can disable network policy too

kqueen/blueprints/api/views.py
naumvd95: no, u in several cases u can manage policy on-the-fly

naumvd95: or i don't understand u , pls clarify info

kqueen/engines/base.py
```diff
@@ -164,6 +164,11 @@ def resize(self, node_count):
         """
         raise NotImplementedError
 
+    def set_network_policy(self, network_provider, enabled):
+        """
+        """
```
katyafervent: Please, provide info at docstring or remove it at all

kqueen/blueprints/api/views.py
katyafervent: Please, add a separate check for application/json above and remove isinstance check at all

if not request.json:
        abort(400, description='JSON data expected')

kqueen/engines/gce.py
```diff
@@ -136,15 +149,35 @@ def _get_client(self):
 
         return client
 
+    def _set_addon_config(self, cluster_config, addon, disabled):
+        """
+
+        """
```
katyafervent: docstring is empty again

kqueen/engines/gce.py
```diff
@@ -189,12 +236,81 @@ def resize(self, node_count, **kwargs):
...
+        """ Updating Network policy configuration on cluster
+        Args:
+            network_provider (str): Name of supported network provider.
+            Currently GKE support only Calico driver.
+
+            enabled (bool): Enabled preferred policy, and deploy additional
+            k8s services, related to network managment.
+
+        Returns:
+             (bool): Result of policy-update
```
katyafervent: you are returning a tuple, please describe it

kqueen/engines/gce.py
```diff
@@ -97,6 +102,13 @@ class GceEngine(BaseEngine):
                 'validators': {
                     'required': True
                 }
+            },
+            'calico_policy': {
```
ekhomyakova:
```
            'network_policy': {
                'type': 'select',
                'label': 'Network Policy',
                'order': 4,
                'choices': [
                    ('none', '(None)'),
                    ('CALICO', 'Calico')
                ],
                'validators': {
                    'required': True
                },
                'class_name': 'network-policy'
            }
```

kqueen/engines/gce.py
```diff
@@ -118,9 +130,20 @@ def __init__(self, cluster, **kwargs):
                 'initialNodeCount': kwargs.get('node_count', 1),
                 'nodeConfig': {
                     'machineType': kwargs.get('machine_type', 'n1-standard-1')
+                },
+                'addonsConfig': {},
+                'networkPolicy': {
```
ekhomyakova:
```                'networkPolicy': {
                    'provider': kwargs.get('network_policy', 'none'),
                    'enabled': bool(kwargs.get('network_policy', False))
                }```

File: kqueen/engines/gce.py [hunk 22]
```diff
@@ -118,9 +130,20 @@ def __init__(self, cluster, **kwargs):
...
+                    'provider': kwargs.get('network_provider', 'CALICO'),
+                    'enabled': kwargs.get('calico_policy', False)
                 }
             }
         }
+        if kwargs.get('calico_policy', False) is True:
```
ekhomyakova: ```        if kwargs.get('network_policy', False):```

kqueen/engines/gce.py
```diff
@@ -158,18 +209,31 @@ def deprovision(self, **kwargs):
         result, error = super(GceEngine, self).deprovision(**kwargs)
         if result:
             return result, error
-
+        request = self.client.projects().zones().clusters().delete(projectId=self.project,
+                                                                   zone=self.zone,
+                                                                   clusterId=self.cluster_id)
         try:
-            self.client.projects().zones().clusters().delete(projectId=self.project, zone=self.zone, clusterId=self.cluster_id).execute()
+            request.execute()
```
ekhomyakova:
```            # TODO: check if provisioning response is healthy
            request.execute()

            policy = self.cluster_config['cluster']['networkPolicy']
            if policy['provider'] != 'none':
                self.cluster.metadata['network_policy'] = policy
                self.cluster.save()
```

kqueen/blueprints/api/views.py
```diff
@@ -220,6 +220,30 @@ def cluster_resize(pk):
...
+@jwt_required()
+def cluster_set_network_policy(pk):
+    obj = get_object(Cluster, pk, current_identity)
+
+    if not request.json:
+        abort(400, description='JSON data expected')
+
+    data = request.json
+    if not all(k in data for k in ('provider', 'enabled')):
+        msg = 'Failed to get network policy configuration'
```
katyafervent: Here we can notify user what is wrong

kqueen/engines/gce.py
```diff
@@ -143,26 +143,38 @@ def __init__(self, cluster, **kwargs):
         self.project = self.service_account_info.get('project_id', '')
         self.zone = kwargs.get('zone', '-')
         self.cluster_id = 'a' + self.cluster.id.replace('-', '')
+
+        # Generate metadata for Network Policies if empty
+        if not isinstance(cluster.metadata.get('network_policy', None), dict):
+            self.cluster.metadata['network_policy'] = network_meta = {}
+            network_meta['provider'] = kwargs.get('network_policy', 'PROVIDER_UNSPECIFIED')
+            network_meta['enabled'] = True if network_meta['provider'] != 'PROVIDER_UNSPECIFIED' \
```
ekhomyakova: network_meta['enabled'] = network_meta['provider'] != 'PROVIDER_UNSPECIFIED'

kqueen/engines/gce.py
```diff
@@ -143,26 +143,38 @@ def __init__(self, cluster, **kwargs):
         self.project = self.service_account_info.get('project_id', '')
         self.zone = kwargs.get('zone', '-')
         self.cluster_id = 'a' + self.cluster.id.replace('-', '')
+
+        # Generate metadata for Network Policies if empty
+        if not isinstance(cluster.metadata.get('network_policy', None), dict):
```
ekhomyakova: `cluster.metadata.get('network_policy')`
None will be by default

kqueen/engines/gce.py
```diff
@@ -143,26 +143,38 @@ def __init__(self, cluster, **kwargs):
...
+            self.cluster.metadata['network_policy'] = network_meta = {}
+            network_meta['provider'] = kwargs.get('network_policy', 'PROVIDER_UNSPECIFIED')
+            network_meta['enabled'] = True if network_meta['provider'] != 'PROVIDER_UNSPECIFIED' \
+                else False
+            logger.debug('Generate metadata for network policies:{}'
```
ekhomyakova: `policies: {}`

kqueen/engines/gce.py
```diff
@@ -143,26 +143,38 @@ def __init__(self, cluster, **kwargs):
...
             'cluster': {
                 'name': self.cluster_id,
                 'initialNodeCount': kwargs.get('node_count', 1),
                 'nodeConfig': {
                     'machineType': kwargs.get('machine_type', 'n1-standard-1')
                 },
-                'addonsConfig': {},
+                'addonsConfig': {
+                    'networkPolicyConfig': {
+                        'disabled': False if meta['network_policy'].
```
ekhomyakova: `'disabled': meta['network_policy'].get('provider', 'PROVIDER_UNSPECIFIED') == 'PROVIDER_UNSPECIFIED'`

kqueen/engines/gce.py
```diff
@@ -143,26 +143,38 @@ def __init__(self, cluster, **kwargs):
...
+        if not isinstance(cluster.metadata.get('network_policy', None), dict):
+            self.cluster.metadata['network_policy'] = network_meta = {}
+            network_meta['provider'] = kwargs.get('network_policy', 'PROVIDER_UNSPECIFIED')
+            network_meta['enabled'] = True if network_meta['provider'] != 'PROVIDER_UNSPECIFIED' \
+                else False
+            logger.debug('Generate metadata for network policies:{}'
+                         .format(cluster.metadata['network_policy']))
+            self.cluster.save()
+
+        meta = self.cluster.metadata
```
ekhomyakova: Looks like you use it only to get network policy, so you can declare `self.cluster.metadata['network_policy']`

kqueen/engines/gce.py
```diff
@@ -221,14 +210,14 @@ def provision(self, **kwargs):
             return False, msg
 
         cluster_config = self.cluster_config['cluster']
-        if cluster_config['networkPolicy']['provider'] is not None:
-            self.cluster.metadata['network_policy'] = cluster_config['networkPolicy']
-            logger.critical('Provisioning cluster {} started, updating metadata...{}'
-                            .format(self.cluster_id, cluster_config['networkPolicy']))
+        meta = self.cluster.metadata
```
ekhomyakova: Same, add `['networkPolicy']`

kqueen/engines/gce.py
```diff
@@ -221,14 +210,14 @@ def provision(self, **kwargs):
...
+        if cluster_config['networkPolicy']['provider'] != 'PROVIDER_UNSPECIFIED':
+            meta['network_policy']['provider'] = cluster_config['networkPolicy']['provider']
```
ekhomyakova: How about
```
meta['network_policy'] = {
    'provider': cluster_config['networkPolicy']['provider'],
    'enabled': cluster_config['networkPolicy']['enabled']
}
```

kqueen/blueprints/api/views.py
naumvd95: done

kqueen/engines/gce.py
naumvd95: done

kqueen/engines/gce.py
naumvd95: done

kqueen/engines/gce.py
naumvd95: done

kqueen/engines/gce.py
naumvd95: done

kqueen/engines/gce.py
naumvd95: in future i prefer to use that for another cases

kqueen/engines/gce.py
naumvd95: done

kqueen/engines/gce.py
naumvd95: done

naumvd95: seems flake8 doesnot apply that

kqueen/engines/gce.py
```diff
@@ -140,14 +194,36 @@ def provision(self, **kwargs):
...
         except Exception as e:
-            msg = 'Creating cluster {} failed with following reason:'.format(self.cluster_id)
+            msg = 'Creating cluster {} failed with the following reason: {}'.format(
+                self.cluster_id, e)
             logger.exception(msg)
             return False, msg
 
+        if cluster_config['networkPolicy']['provider'] != 'PROVIDER_UNSPECIFIED':
+            network_meta['provider'] = cluster_config['networkPolicy']['provider']
+            network_meta['provider'] = cluster_config['networkPolicy']['provider']
```
ekhomyakova: provider -> enabled

kqueen/config/base.py
```diff
@@ -29,7 +29,7 @@ class BaseConfig:
     CLUSTER_OK_STATE = 'OK'
     CLUSTER_PROVISIONING_STATE = 'Deploying'
     CLUSTER_DEPROVISIONING_STATE = 'Destroying'
-    CLUSTER_RESIZING_STATE = 'Resizing'
+    CLUSTER_UPDATING_STATE = 'Updating'
```
atengler: You need to do corresponding change in the UI as well, so both API and UI work with the same default states

naumvd95: done

kqueen/engines/gce.py
```diff
@@ -158,18 +236,31 @@ def deprovision(self, **kwargs):
...
+                                                                                repr(e))
             logger.exception(msg)
             return False, msg
 
         return True, None
 
     def resize(self, node_count, **kwargs):
+
+        if int(node_count) < 2 and \
+           self.cluster_config['cluster']['networkPolicy']['enabled'] is True:
```
atengler: Seems like incorrect indentation

naumvd95: everything is ok, flake passed, travis passed)

PR #281, opened 2018-04-27
Add lock on save to etcd operation

katyafervent:
* It prevents situation, when somebody
  creates users with identical names at the same time
* Fix bug with unique field detection for non-global namespaced models
* Add unittest with unique field

### DIFF
--- a/kqueen/storages/etcd.py
+++ b/kqueen/storages/etcd.py
@@ -583,6 +583,7 @@ class Model:
     def save(self, validate=True, assign_id=True):
         """Save object to database
 
+        Hold lock during saving to avoid interruption into unique fields check
 
         Attributes:
             validate (bool): Validate model before saving. Defaults to `True`.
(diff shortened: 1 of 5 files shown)


kqueen/storages/etcd.py
```diff
@@ -592,24 +593,24 @@ def save(self, validate=True, assign_id=True):
             bool: `True` if model was saved without errors, `False` otherwise.
 
         """
+        with etcd.Lock(current_app.db.client, 'customer1'):
+            if assign_id:
+                self.verify_id()
 
-        if assign_id:
-            self.verify_id()
+            validation_status, validation_msg = self.validate()
```
naumvd95: maybe wrap it in try-catch block?

kqueen/storages/etcd.py
```diff
@@ -592,24 +593,24 @@ def save(self, validate=True, assign_id=True):
...
-            current_app.db.client.write(key, self.serialize())
-
-            self._key = key
-            return True
-        except Exception:
-            raise
+                self._key = key
+                return True
+            except Exception:
+                raise
```
naumvd95: what to raise? maybe we should add generic kqueen exception for it?

kqueen/storages/etcd.py
```diff
@@ -637,12 +638,17 @@ def validate(self):
                 return False, 'Required field {} is None'.format(field)
 
             if field_object.unique and field_object.value:
-                for k, v in self.list(self.namespace).items():
+                if self.is_namespaced():
```
naumvd95: btw what do u think about method naming and method itself
imo at least we should rename it as (is_global) or smth similar
coz global -namespace is namespace too, so imo this method is weird

kqueen/storages/etcd.py
katyafervent: I just add with statement here, just the indent is changed. I agree, that we need to add try-catch, but first we need to raise an exception in validate method instead of returning Boolean value

kqueen/storages/etcd.py
katyafervent: I just change the indent in scope of this commit, the catched exception is raised again here, but since we do nothing more, I don't know why it is here

kqueen/storages/etcd.py
katyafervent: May be we should, but not in scope of this commit

PR #282, opened 2018-04-30
Calico aws

naumvd95:
WIP Define parameters  overriding for Jenkins

    Use cases:
    1) Calico managing in case of using mcp-job calico-ha-aws-k8s
    2) overriding parameters in case of any jenkins job

    From UI part:
    At least need to provide possibility of defining key(str):value(str) parameters
scheme = https://www.lucidchart.com/invitations/accept/6e2bbc84-a5df-43f9-8b21-2a5b754a57d8

### DIFF
--- a/kqueen/engines/jenkins.py
+++ b/kqueen/engines/jenkins.py
@@ -48,8 +48,37 @@ class JenkinsEngine(BaseEngine):
                     'required': True
                 }
             }
+            'override_parameters': {
+                'type': 'text',
+                'label': 'Key=Value line by line',
+                'order': 2,
+            },
         },
-        'cluster': {}
+        'cluster': {
+            'network_policy': {
+                'type': 'select',
+                'choices': [
+                    ('calico', 'Calico Network Provider'),
+                ],
+                'label': 'Network Policy',
+                'order': 3,
+             },
+            'enable_policy': {
+                'type': 'checkbox',
+                'label': 'Network Policy',
+                'order': 4,
+                'checkbox_text': 'Enable Network Policy (2 or more nodes are required)',
+                'class_name': 'nwpolicy-checkbox'
+             }
+            'network_range': {
+                'type': 'string',
+                'label': 'Network Range',
+                'default': 5,
+                'validators': {
+                #TODO CIDR validation
+                }
+            }
+        }
     }
 
     def __init__(self, cluster, **kwargs):
@@ -64,8 +93,28 @@ class JenkinsEngine(BaseEngine):
         self.client = self._get_client()
         # Cache settings
         self.cache_timeout = 5 * 60
+        self.cluster_config = {
+            'cluster': {
+                'name': self.cluster_id,
+                'networkPolicy': {
+                    'provider': kwargs.get('network_policy', 'calico')
+                    'enabled': kwargs.get('enable_policy', False)
+                    'network_range': kwargs.get('network_range', {})
+                }
+            }
+        }
+        self.job_override_params = kwargs.get('override_parameters', [])
 
     def _get_provision_job_builds(self):
+        """
+        Get builds history of Jenkins job used to provision clusters
+
+        Returns:
+            dict: More information at :func:`~jenkins.Jenkins.get_job_info`
+        """
+        return self.client.get_job_info(self.provision_job_name, depth=1)
+
+    def _parse(self):
         """
         Get builds history of Jenkins job used to provision clusters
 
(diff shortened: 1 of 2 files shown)


kqueen/engines/jenkins.py
```diff
@@ -64,6 +81,21 @@ def __init__(self, cluster, **kwargs):
         self.client = self._get_client()
         # Cache settings
         self.cache_timeout = 5 * 60
+        self.job_override_params = kwargs.get('override_parameters', {})
+        logger.critical('OVERRIDEDEEE {}'.format(self.job_override_params))
```
katyafervent: :sweat_smile:

kqueen/engines/jenkins.py
```diff
@@ -74,39 +106,40 @@ def _get_provision_job_builds(self):
...
-        try:
-            client = jenkins.Jenkins(config.get('JENKINS_API_URL'), **conn_kw)
-            auth_verify = client.get_whoami()
-            if auth_verify:
-                status = config.get('PROVISIONER_OK_STATE')
-        except Exception as e:
-            logger.exception('Could not contact JenkinsEngine backend: ')
-            status = config.get('PROVISIONER_ERROR_STATE')
-        return status
+        Check, that defined key parameter exist in Job
```
katyafervent: that defined key parameter exist in Job -> that key parameter, defined by the user exists in the specified Job

kqueen/engines/jenkins.py
```diff
@@ -74,39 +106,40 @@ def _get_provision_job_builds(self):
...
-            :obj:`jenkins.Jenkins`: initialized Jenkins client
+            bool: True if all parameters exist
         """
-        return jenkins.Jenkins(self.jenkins_url, **{
-            'username': self.username,
-            'password': self.password,
-            'timeout': 10
-        })
+
+        job_body = self.client.get_job_info(self.provision_job_name, depth=1)
```
katyafervent: I suggest to add try-catch here and return False from function end at catch block

kqueen/engines/jenkins.py
```diff
@@ -118,6 +151,17 @@ def provision(self, **kwargs):
         # PATCH THE CTX TO CONTAIN CLUSTER NAME AND UUID
         ctx[cluster_name] = 'kqueen-{}'.format(self.cluster.id)
         ctx[cluster_uuid] = self.cluster.id
+        if self.job_override_params:
```
katyafervent: I suggest to manage all job parameters in a one place, so first let's add cluster_name and cluster_uuid to self.job_override_params (also for me self.job_parameters sounds better) and then manage everything in the same way

kqueen/engines/jenkins.py
```diff
@@ -118,6 +151,17 @@ def provision(self, **kwargs):
...
+            for key, value in self.job_override_params:
```
katyafervent: self.job_override_params.items()

kqueen/engines/jenkins.py
```diff
@@ -118,6 +151,17 @@ def provision(self, **kwargs):
...
+                ctx[key] = ctx[value]
+
+        for param in ctx.keys():
+            if not self._parameter_exist(param):
+                msg = 'Failed to manage job parameter: {}'.format(param)
+                logger.error(msg)
+                return False, msg
+
+        logger.debug('Current Job configuration: {}, provisioning started...'.format(ctx))
```
katyafervent: actually, it is started after 166 line

kqueen/engines/jenkins.py
naumvd95: imo its different cases - override-parameters used for overriding sane-default parameters, like boolean variables, checkers, additional cases
cluster name and UUID - must have cases, and it doesnot have any sane-defaults, so i prefer to manage it from separate UI-fields

kqueen/engines/jenkins.py
```diff
@@ -47,9 +47,26 @@ class JenkinsEngine(BaseEngine):
                 'validators': {
                     'required': True
                 }
+            },
+            # TODO below form should be increased dynamically in case of 'add one more parameter'
+            # Use-case: 'add parameter'->'key': 'value'-> 'add parameter'-> etc...
+            # Should return dict{}
+            'override_parameters': {
```
katyafervent: please, replace with
            'override_parameters': {
                'type': 'parameters',
                'label': 'Jenkins job parameters',
                'order': 2,
                'validators': {
                          'required': False
                }
            }
        },
        'cluster': {}
    }

Also, did you delete clusters occasionally?

kqueen/engines/jenkins.py
naumvd95: there are try catch block, so i prefer to log here, its bad?

katyafervent: it's better to place it inside try-catch block

naumvd95:
closed as duplicate

PR #293, opened 2018-05-11
Prepare release 1.0

naumvd95:
* Search mechanism for Ldap case (OU is no longer required)
* Calico networking support for GKE engine
* Openstack engine based on Kubespray 
* Lock on save to etcd operations
* Huge documentation refactoring (on readthedocs.io as well)
* Bugfixes
* Minor infra changes

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@
 
 from setuptools import setup, find_packages
 
-version = '0.19'
+version = '1.0'
 
 with open('README.rst') as f:
     long_description = ''.join(f.readlines())


ekhomyakova:
OU no longer required -> OU is no longer required
and on readthedocs.io too -> on readthedocs.io as well

docker-compose.demo.yml
```diff
@@ -1,7 +1,7 @@
 version: '3.2'
 services:
   api:
-    image: kqueen/api:v0.19
+    image: kqueen/api:v1.0
```
atengler: I would not do this as part of the PR, just do the `setup.py` change. You need to merge this PR and tag the merge commit as `v1.0` so the image can get built, it won't exist until you push the tag, so I would create separate PR with the demo file update after the images are built already and you can test them

PR #306, opened 2018-05-21
Add 'ERROR' status to GCE provisioner

katyafervent:
* Add forgotten ERROR status
* Add provided by google API error description to the cluster metadata

### DIFF
--- a/kqueen/engines/gce.py
+++ b/kqueen/engines/gce.py
@@ -213,10 +213,9 @@ class GceEngine(BaseEngine):
             request.execute()
             # TODO: check if provisioning response is healthy
         except Exception as e:
-            msg = 'Creating cluster {} failed with the following reason: {}'.format(self.cluster_id,
-                                                                                    e)
+            msg = 'Creating cluster {} failed with the following reason: {}'.format(self.cluster_id, e)
             logger.exception(msg)
-            return False, msg
+            return False, e
 
         if cluster_config['networkPolicy']['provider'] != 'PROVIDER_UNSPECIFIED':
             network_meta['provider'] = cluster_config['networkPolicy']['provider']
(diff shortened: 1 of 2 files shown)


kqueen/engines/gce.py
```diff
@@ -429,7 +429,7 @@ def cluster_get(self):
             'name': self.cluster_id,
             'id': self.cluster.id,
             'state': state,
-            'metadata': {}
+            'metadata': {'status_message': response['statusMessage']} if 'statusMessage' in response else {}
```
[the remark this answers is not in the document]
katyafervent: Maybe you'll find better or ask me before disapproving the PR? https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.zones.clusters 
'statusMessage' is in output only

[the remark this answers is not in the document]
naumvd95: ill request changes just to freeze PR) ok thanks for clarifying

naumvd95:
also tests are failed for that PR

kqueen/engines/gce.py
ekhomyakova: `response['statusMessage']} if 'statusMessage' in response else {}`
->
`response.get('statusMessage', {})`

kqueen/models.py
```diff
@@ -62,6 +64,13 @@ def get_state(self):
 
         return self.state
 
+    def set_status(self, cluster):
+        if cluster:
```
ekhomyakova: What do you expect to get that will not pass this condition? Manually passed None or empty string?

katyafervent: in case of fail cluster is empty

kqueen/engines/gce.py
katyafervent: No, I need empty dictionary or dictionary with the 'status_message' key

kqueen/models.py
ekhomyakova: Empty dict? If so, you don't have to check it here, because you check it below

kqueen/engines/gce.py
ekhomyakova: Oh, didn't notice, okay

PR #309, opened 2018-05-25
Make organization namespace unique

katyafervent:
* Organization namespace corresponds to namespace in k8s,
  so it's not allowed to have idential names.
* Fix corresponding tests, before tests create several organization
  with same namespace

### DIFF
--- a/kqueen/blueprints/api/test_organization.py
+++ b/kqueen/blueprints/api/test_organization.py
@@ -36,6 +36,13 @@ class TestOrganizationCRUD(BaseTestCRUD):
             assert response.status_code == 500
             assert isinstance(remaining, list) and remaining
 
+    def get_create_data(self):
+        data = self.obj.get_dict(expand=True)
+        data.update(self.get_edit_data())
+        data['id'] = None
+
+        return data
+
     def test_policy(self):
         url = url_for('api.organization_policy', pk=self.obj.id)
 
--- a/kqueen/conftest.py
+++ b/kqueen/conftest.py
@@ -11,11 +11,12 @@ import etcd
 import faker
 import json
 import pytest
+import random
+import string
 import uuid
 import yaml
 
 config = current_config()
-fake = faker.Faker()
 fake = faker.Faker()
 current_app = None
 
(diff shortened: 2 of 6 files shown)


naumvd95:
1 notice, pls fix tests

kqueen/blueprints/api/test_organization.py
```diff
@@ -36,6 +36,13 @@ def test_crud_delete(self):
             assert response.status_code == 500
             assert isinstance(remaining, list) and remaining
 
+    def get_create_data(self):
```
ekhomyakova: Unclear naming. Did you mean "creation"?

kqueen/storages/etcd.py
```diff
@@ -650,8 +650,9 @@ def validate(self):
                     if getattr(v, field) == field_object.value:
                         return False, 'Field "{name}" should be unique'.format(name=field)
 
-            if field_object.value and not field_object.validate():
-                return False, 'Field {} validation failed'.format(field)
+            if field_object.value is not None:
+                if not field_object.validate():
```
ekhomyakova: Why don't you use `and` with the previous `if`?

kqueen/conftest.py
```diff
@@ -103,14 +104,18 @@ def destroy(self):
...
+            name = ''.join(random.choice(string.ascii_letters) for m in range(8))
         organization = Organization(
             None,
-            name='DemoOrg',
-            namespace='demoorg',
+            name=name,
+            namespace=name
         )
-        organization.save()
+        # if name not in [d.name for d in Organization.list(name).values()]:
```
ekhomyakova: Looks like smth that should be removed or uncommented

kqueen/conftest.py
```diff
@@ -103,14 +104,18 @@ def destroy(self):
 
 
 class OrganizationFixture:
-    def __init__(self):
+    def __init__(self, name=None):
         """Prepare organization object."""
+        if not name:
+            name = ''.join(random.choice(string.ascii_letters) for m in range(8))
```
ekhomyakova: `for _ in`

kqueen/blueprints/api/test_organization.py
katyafervent: that name defined in base class

PR #319, opened 2018-06-13
Add autopep8

katyafervent:


git_hooks/pre-commit
```diff
@@ -0,0 +1,82 @@
...
+import re
+import subprocess
+import sys
+
+# Use just one of these
+select_codes = []
+ignore_codes = []
+
+# Add parameters like '--max-line-length' below
+overrides = ['--max-line-length=120']
```
ekhomyakova: 100

setup.cfg
```diff
@@ -8,8 +8,7 @@ env =
 	D:prometheus_multiproc_dir=/tmp/prometheus_client/
 
 [flake8]
-ignore = E501
-max-line-length = 160
+max-line-length = 120
```
ekhomyakova: 100

katyafervent: In that case all files need to be refactored, why do we need exact 100?

ekhomyakova: We can use 79 if you want. 100 is a compromise because it's a standard for JS files (which we have enough in the project), so we can use the same max length for all the files. 
Why do we need 120?

naumvd95:
IMO there are several ways to resolve this problem:
1) disable check
2) disable check & create US for refactoring to 100 symbols if anybody have time for it
3) leave 120 symbols checker as is, coz imo 120 checker is better than nothing

ekhomyakova:
> Hooks can be muted with --no-verify option:  ``git commit --no-verify``
Are we able to merge the code without refactoring in this request? And do the separate request where the code changes will be applied by the hooks

katyafervent:
@ekhomyakova I disable max length check again. hook is directed to the specific commit, so it is impossible to extract changes made by the hook to a separate commit

PR #321, opened 2018-06-18
Heptio Auth configuration

naumvd95:
Kqueen-aws-heptio, Issue  in official python k8s-client :
https://github.com/kubernetes-client/python/issues/514
imo it useless to do hardcode (bash scripts/get token inside docker, override k8s-config on the fly) without 'I_WANT_IT_NOW' request?
if we dont hurry, we may wait due issue-resolving
full topic - https://github.com/kubernetes/kubernetes/issues/62185

### DIFF
--- a/docs/kqueen.rst
+++ b/docs/kqueen.rst
@@ -415,7 +415,15 @@ Provision a Kubernetes cluster using Amazon Elastic Kubernetes Service
 #. Create your Amazon EKS service role, Amazon EKS Cluster VPC - `Official EKS Quickstart <https://docs.aws.amazon.com/eks/latest/userguide/getting-started.html>`_.
 #. Create Amazon EKS IAM Policy - `Creating Amazon EKS IAM Policies <https://docs.aws.amazon.com/eks/latest/userguide/EKS_IAM_user_policies.html>`_.
 #. Create Amazon EKS IAM Passrole Policy - `Creating Amazon Passrole Policies <https://aws.amazon.com/blogs/security/granting-permission-to-launch-ec2-instances-with-iam-roles-passrole-permission/>`_. In policy json-snippet switch all ``ec2`` entities to ``eks``.
-#. Create User account with Programmatic Access and attach all created policies - `Creating an IAM User in Your AWS Account <https://docs.aws.amazon.com/IAM/latest/UserGuide/id_users_create.html>`_. Don't forget to save User Key/Secret Pair.
+#. Create User account with Programmatic Access and attach created policies - `Creating an IAM User in Your AWS Account <https://docs.aws.amazon.com/IAM/latest/UserGuide/id_users_create.html>`_. Don't forget to save User Key/Secret Pair.
+#. Attach additional policies to User account for Heptio Authentication:
+
+  #. AmazonEC2FullAccess
+  #. AmazonRoute53FullAccess
+  #. AmazonS3FullAccess
+  #. IAMFullAccess
+  #. AmazonVPCFullAccess
+
 #. Log in to the KQueen web UI.
 #. From the ``Create Provisioner`` tab, select ``Amazon Elastic Kubernetes Service`` and set the
    following:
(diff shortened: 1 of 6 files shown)


kqueen/engines/eks.py
```diff
@@ -70,18 +70,18 @@ class EksEngine(BaseEngine):
...
-                'label': 'Node Count',
-                'default': 3,
-                'order': 3,
-                'validators': {
-                    'required': True,
-                    'min': 1,
-                    'number': True
-                }
-            },
+#            'node_count': {
```
katyafervent: I also suggest to here add link  to the issue, that we are waiting to be resolved

kqueen/engines/eks.py
```diff
@@ -206,17 +206,21 @@ def get_kubeconfig(self):
...
-            self.cluster.kubeconfig["users"][0]["user"]["exec"]["args"][2] = cluster['cluster']['name']
+            self.cluster.kubeconfig['users'][0]['user']['exec']['args'][2] = cluster['cluster']['name']
+            # Set user credentials for Heptio auth
+            self.cluster.kubeconfig['users'][0]['user']['exec']['env'][0]['value'] = self.aws_access_key
+            self.cluster.kubeconfig['users'][0]['user']['exec']['env'][1]['value'] = self.aws_secret_key
+
             # TODO define , do we need to use user Arn or Cluster Arn
-#            self.cluster.kubeconfig["users"][0]["user"]["exec"]["args"][4] = cluster['cluster']['roleArn']
-#            self.cluster.kubeconfig["users"][0]["user"]["exec"]["args"][4] = cluster['cluster']['arn']
+#            self.cluster.kubeconfig['users'][0]['user']['exec']['args'][4] = cluster['cluster']['roleArn']
```
katyafervent: and here at TODO

entrypoint.sh
```diff
@@ -1,5 +1,14 @@
 #!/bin/bash
+set -ex
 
+# install aws dependencies
+curl -o heptio-authenticator-aws https://amazon-eks.s3-us-west-2.amazonaws.com/1.10.3/2018-06-05/bin/linux/amd64/heptio-authenticator-aws
+chmod +x ./heptio-authenticator-aws
+mkdir -p $HOME/bin
+cp ./heptio-authenticator-aws $HOME/bin/heptio-authenticator-aws && export PATH=$HOME/bin:$PATH
+echo 'export PATH=$HOME/bin:$PATH' >> ~/.bashrc
```
katyafervent: I suggest to install aws dependencies in a separate file, so if some fail occurred, exporting variables will work, what do you think?

docs/kqueen.rst
```diff
@@ -436,6 +444,13 @@ Provision a Kubernetes cluster using Amazon Elastic Kubernetes Service
 #. Click ``Submit``.
 #. To track the cluster status, navigate to the KQueen main dashboard.
 
+.. note::
+
+   Currently Kqueen does not support Kubernetes workers-management due Amazon EKS API restrictions,
```
katyafervent: Currently, KQueen does not support Kubernetes workers-management because of the Amazon EKS API restrictions. To cover the Kubernetes workers use cases, see `Getting started with Amazon EKS <https://docs.aws.amazon.com/eks/latest/userguide/getting-started.html>`_

kqueen/engines/eks.py
naumvd95: seems there are no link, we have not got api support of cluster resizing

kqueen/engines/eks.py
naumvd95: ok

kqueen/engines/eks.py
```diff
@@ -70,18 +70,18 @@ class EksEngine(BaseEngine):
...
-                'default': 3,
-                'order': 3,
-                'validators': {
-                    'required': True,
-                    'min': 1,
-                    'number': True
-                }
-            },
+#            'node_count': {
+#              TODO  currently unsupported due EKS API restrictions
```
ekhomyakova: due to?
https://en.oxforddictionaries.com/definition/due
https://en.oxforddictionaries.com/definition/due_to

PR #325, opened 2018-06-27
Return correct codes an fix bugs

katyafervent:


kqueen/auth/common.py
```diff
@@ -140,11 +140,12 @@ def identity(payload):
 
 
 def encrypt_password(_password):
-    config = current_config()
-    rounds = config.get('BCRYPT_ROUNDS', 12)
-    password = str(_password).encode('utf-8')
-    encrypted = bcrypt.hashpw(password, bcrypt.gensalt(rounds)).decode('utf-8')
-    return encrypted
+    if _password:
```
naumvd95: why we need this?
seems previous case failed with valid traceback in case of empty pass

kqueen/blueprints/api/generic_views.py
```diff
@@ -278,6 +278,11 @@ def dispatch_request(self, *args, **kwargs):
             try:
                 self.save_object()
                 self.after_save()
+            except ValueError as e:
+                if 'unique' in str(e):
```
naumvd95: IMO its bad case to manage error codes by string-checking, traceback can change when package updates
and IMO both this cases can be managed by 400 code

kqueen/models.py
```diff
@@ -450,4 +450,4 @@ def namespace(self):
         Returns:
             str: Namespace (from organization)
         """
-        return self.organization.namespace
+        return self.organization.namespace if self.organization else None
```
naumvd95: Maybe we remove this method? it contains ~10 lines, but useful part is ```self.organization.namespace```

kqueen/auth/common.py
katyafervent: encrypted successfully generated in case of None

kqueen/blueprints/api/generic_views.py
katyafervent: ok, I'll think how it will be better to handle that

kqueen/models.py
katyafervent: It is a property so that all objects will are consistent and have namespace attribute

kqueen/auth/common.py
katyafervent: and you still get encrypted password for None

naumvd95: maybe its better to deny None-passwords at all?
i mean like that
```+++ b/kqueen/blueprints/api/views.py
@@ -512,7 +512,7 @@ def user_password_update(pk):
     if not isinstance(data, dict):
         abort(400)

-    obj.password = encrypt_password(data.get('password'))
+    obj.password = encrypt_password(data['password'])

     try:
         obj.save()```

Comment (naumvd95) [reply kqueen/auth/common.py]: your fix is good, but seems problem is deeper

Comment (katyafervent) [reply kqueen/auth/common.py]: you'll get key error and request will not finish as expected, now 400 - bad request is returned during modal validation - password is None but it is required in Model

Comment (naumvd95) [reply kqueen/auth/common.py]: what about that?
```
diff --git a/kqueen/blueprints/api/views.py b/kqueen/blueprints/api/views.py
index cc63da4..69194a4 100644
--- a/kqueen/blueprints/api/views.py
+++ b/kqueen/blueprints/api/views.py
@@ -512,6 +512,9 @@ def user_password_update(pk):
     if not isinstance(data, dict):
         abort(400)

+    if data.get('password') is None:
+        abort(400, description='Password required')
+
     obj.password = encrypt_password(data.get('password'))

     try:
(END)
```

PR #327, opened 2018-06-28
Increase verbosity of Exception Flash messages

naumvd95:


kqueen/engines/aks.py
```diff
@@ -197,12 +197,14 @@ def provision(self, **kwargs):
         )
 
         try:
-            self.client.managed_clusters.create_or_update(self.resource_group_name, self.cluster.id, managed_cluster)
+            self.client.managed_clusters.create_or_update(self.resource_group_name,
+                                                          self.cluster.id, managed_cluster)
```
katyafervent: if you selected that style, managed_cluster should be in the new line also

kqueen/engines/aks.py
```diff
@@ -238,10 +241,12 @@ def resize(self, node_count, **kwargs):
...
             # TODO: check if resizing response is healthy
         except Exception as e:
-            msg = 'Resizing cluster {} failed with following reason:'.format(self.cluster.id)
+            msg = 'Resizing cluster {} failed with following reason: {}'\
```
katyafervent: if you are logging exception e will be logged anyway, we just need to return repr(e) instead of msg here

kqueen/engines/aks.py
```diff
@@ -219,9 +221,10 @@ def deprovision(self, **kwargs):
             self.client.managed_clusters.delete(self.resource_group_name, self.cluster.id)
             # TODO: check if deprovisioning response is healthy
         except Exception as e:
-            msg = 'Deleting cluster {} failed with following reason:'.format(self.cluster.id)
+            msg = 'Deleting cluster {} failed with following reason: '\
```
ekhomyakova: Why did you break this line but didn't break L205? They are of the same length

kqueen/engines/aks.py
```diff
@@ -197,12 +197,14 @@ def provision(self, **kwargs):
...
 
         try:
-            self.client.managed_clusters.create_or_update(self.resource_group_name, self.cluster.id, managed_cluster)
+            self.client.managed_clusters.create_or_update(self.resource_group_name,
+                                                          self.cluster.id,
+                                                          managed_cluster)
             # TODO: check if provisioning response is healthy
         except Exception as e:
-            msg = 'Creating cluster {} failed with following reason {}:'.format(self.cluster.id, e)
+            msg = 'Creating cluster {} failed with following reason:'.format(self.cluster.id)
```
ekhomyakova: *the following

kqueen/engines/aks.py
```diff
@@ -238,12 +241,15 @@ def resize(self, node_count, **kwargs):
...
 
         try:
-            self.client.managed_clusters.create_or_update(self.resource_group_name, self.cluster.id, managed_cluster)
+            self.client.managed_clusters.create_or_update(self.resource_group_name,
+                                                          self.cluster.id,
+                                                          managed_cluster)
             # TODO: check if resizing response is healthy
         except Exception as e:
-            msg = 'Resizing cluster {} failed with following reason:'.format(self.cluster.id)
+            msg = 'Resizing cluster {} failed with following reason: '\
```
ekhomyakova: *the following

PR #326, opened 2018-06-28
Improve azure provisioner: install specific packages, update field names and fix cluster resize

katyafervent:
Often azure brakes package dependencies and our CI became broken.
Install only that azure packages, that we are using.

PR #328, opened 2018-06-28
Hotfix for avoid travis issues

naumvd95:
- Resizing works  fine
Please double-check before merge
need to merge that asap, then work on https://github.com/Mirantis/kqueen/pull/326

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ setup(
         'google-api-python-client==1.6.4',
         'google-auth==1.2.1',
         'google-auth-httplib2==0.0.3',
-        'azure',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -50,7 +50,7 @@
         'google-api-python-client==1.6.4',
         'google-auth==1.2.1',
         'google-auth-httplib2==0.0.3',
-        'azure',
+        'azure==2.0.0',
```
katyafervent: it is the old version, change to 3.0.0 please

naumvd95: 3.0.0 doesnot work properly , pls check by yourself

PR #329, opened 2018-06-29
Support clusters/provisioners filtering

ekhomyakova:
Related: https://github.com/Mirantis/kqueen-ui/pull/105

### DIFF
--- a/kqueen/blueprints/api/generic_views.py
+++ b/kqueen/blueprints/api/generic_views.py
@@ -192,7 +192,30 @@ class ListView(GenericView):
         else:
             self.obj = sorted_by_date
 
+    def _filter_objects(self, objects, filters):
+        if not filters:
+            return objects
+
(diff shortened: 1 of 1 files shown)


kqueen/blueprints/api/generic_views.py
```diff
@@ -192,7 +192,29 @@ def _save_objects_range(self, objects):
...
+        if not filters:
+            return objects
+        return list(filter(
+            lambda x: all(filters[key].lower() in get_name(x, key).lower()
+                          for key in filters.keys()),
+            objects
+        ))
+
     def set_object(self, *args, **kwargs):
+        supported_filters = ['name', 'provisioner']
```
naumvd95: IMO it will be better to rename 'name' to `cluster_name`

ekhomyakova: It's not only for clusters

PR #330, opened 2018-06-29
Fix deprovisioning PVC in kubespray engine

redixin:
Volumes created by k8s cluster should be deleted along with cluster itself

### DIFF
--- a/kqueen/engines/openstack_kubespray.py
+++ b/kqueen/engines/openstack_kubespray.py
@@ -229,7 +229,7 @@ class OpenstackKubesprayEngine(BaseEngine):
             pvc_names = self._cleanup_pvc()
             self.ks.delete()
         except Exception as e:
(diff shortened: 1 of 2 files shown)


kqueen/engines/openstack_kubespray.py
```diff
@@ -224,16 +226,37 @@ def _run_provisioning(self):
 
     def deprovision(self):
         try:
+            pvc_names = self._cleanup_pvc()
             self.ks.delete()
         except Exception as e:
             logger.warn("Unable to cleanup cluster data: %s" % e)
+            self.cluster.state = config.CLUSTER_ERROR_STATE
```
naumvd95: afaik we need to use logger exception here^^^

naumvd95:
@redixin take a look on comment and rebase manually before merge , coz branch outdated

PR #331, opened 2018-07-04
Stop provisioning in case of error

katyafervent:
Also, error is saved in 'status_message' field
and the reason of error will be avaliable in UI

### DIFF
--- a/kqueen/engines/openstack_kubespray.py
+++ b/kqueen/engines/openstack_kubespray.py
@@ -78,7 +78,7 @@ class OpenstackKubesprayEngine(BaseEngine):
                 "help_message": "Must be odd number",
                 "validators": {
                     "required": True,
-                    "minimum": 3,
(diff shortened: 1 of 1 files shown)


kqueen/engines/openstack_kubespray.py
```diff
@@ -196,7 +199,14 @@ def provision(self):
         try:
             self.cluster.state = config.CLUSTER_PROVISIONING_STATE
             self.cluster.save()
-            app.executor.submit(self._run_provisioning)
+            future = app.executor.submit(self._run_provisioning)
```
naumvd95: future? imo its not a good naming)

katyafervent: that is how object called, that ThreadPoolExecutor returns https://docs.python.org/3/library/concurrent.futures.html#executor-objects

PR #333, opened 2018-07-04
Support sorting for clusters/provisioners

ekhomyakova:
If no sorting parameters provided, list will be returned as is, without any sorting.

Related: https://github.com/Mirantis/kqueen-ui/pull/107

### DIFF
--- a/docs/api.yml
+++ b/docs/api.yml
@@ -75,6 +75,19 @@ paths:
               total:
                 type: "int"
 
+   /clusters?sortby={field}&order=asc:
+    get:
+      summary: Get sorted list of clusters (multi-fields sorting is not supported)
+      produces:
+        - "application/json"
+      responses:
+        200:
+          description: "Successful operation"
+          schema:
+            type: "array"
+            items:
+              $ref: "#/definitions/Cluster"
+
   /clusters/{id}:
     get:
       summary: Get information about cluster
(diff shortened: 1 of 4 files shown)


docs/api.yml
```diff
@@ -57,7 +57,7 @@ paths:
             items:
               $ref: "#/definitions/Cluster"
 
-  /clusters?limit={limit}&offset={offset}:
+  /clusters?limit={limit}&offset={offset}&sortby={field}&order=asc:
```
katyafervent: and what about provisioners list? Please add it also, and provide the default way if order is not provided

kqueen/blueprints/api/generic_views.py
```diff
@@ -181,42 +181,49 @@ def dispatch_request(self, *args, **kwargs):
...
     def set_object(self, *args, **kwargs):
-        supported_filters = ['name', 'provisioner']
+        supported_filters = ['name', 'provisioner', 'engine']
         filters = {}
         for filter_name in supported_filters:
             filter_value = request.args.get(filter_name)
             if filter_value:
                 filters[filter_name] = filter_value
 
+        supported_sort_fields = ['name', 'provisioner', 'engine',
```
katyafervent: I suggest to make it a class constant and to convert it to a dictionary (with sorting key as values), so get_sorting_key will not be needed

docs/api.yml
katyafervent: and please add notice, that sorting by multiple keys is not supported

PR #334, opened 2018-07-05
Validate before provision

katyafervent:


kqueen/engines/openstack_kubespray.py
```diff
@@ -561,38 +568,60 @@ def __init__(self, stack_name, *, os_kwargs, cluster, extra_ssh_key):
...
+            ipaddress.ip_address(address)
+            return address
+
+        mc = self.cluster.metadata["master_count"]
+        if mc % 2 == 0 or mc == 1:
+            raise ValueError("Master node count must be an odd number at least 3 or greater")
+        self.meta["master_count"] = mc
+        self.meta["slave_count"] = self.cluster.metadata["slave_count"]
+
+        self.meta["dns"] = [validate_ip(ip) for ip in
```
ekhomyakova: I suggest to place it after all `is None` checks.

katyafervent: why? I placed it here since it performs faster then checks below, so if error will be here, response will be sent faster

ekhomyakova: Why is it faster, I don't get it?

kqueen/engines/openstack_kubespray.py
```diff
@@ -561,38 +568,60 @@ def __init__(self, stack_name, *, os_kwargs, cluster, extra_ssh_key):
...
+
         resources = {
             "masters": [],
             "slaves": [],
         }
         network = self.c.create_network(self.stack_name)
         subnet = self.c.create_subnet(network, cidr="10.1.0.0/16",
                                       subnet_name=self.stack_name,
-                                      dns_nameservers=dns)
+                                      dns_nameservers=self.meta['dns'])
```
naumvd95: maybe get() ?

kqueen/engines/openstack_kubespray.py
```diff
@@ -561,38 +568,60 @@ def __init__(self, stack_name, *, os_kwargs, cluster, extra_ssh_key):
...
         router = self.c.create_router(name=self.stack_name,
-                                      ext_gateway_net_id=ext_net.id)
+                                      ext_gateway_net_id=self.meta['ext_net'].id)
```
naumvd95: maybe get() ?

kqueen/engines/openstack_kubespray.py
```diff
@@ -602,9 +631,9 @@ def provision(self):
                 "hostname": master.name,
             })
         for slave in self._boot_servers(name=self.stack_name,
-                                        servers_range=range(slave_count),
-                                        image=image,
-                                        flavor=flavor,
+                                        servers_range=range(self.meta["slave_count"]),
+                                        image=self.meta['image'],
+                                        flavor=self.meta['flavor'],
```
naumvd95: maybe get() ?

kqueen/engines/openstack_kubespray.py
katyafervent: since it doesn't send api call to the openstack

ekhomyakova: Oh, okay

kqueen/engines/openstack_kubespray.py
katyafervent: Some fields are required and some are not, let me also send 400 if required field is not provided

naumvd95: okok

PR #336, opened 2018-07-06
Revert "Validate OS-kubespray cluster before provision"

naumvd95:
Reverts Mirantis/kqueen#334


naumvd95:
@katyafervent can we merge it?

PR #337, opened 2018-07-06
Send 400 in case of missed required parameters

katyafervent:
While creating OS-kubespray cluster if some parameters were missed,
500 was sent.
Also use get method for checking avalibility zone,
since it is optional parameter

### DIFF
--- a/kqueen/engines/openstack_kubespray.py
+++ b/kqueen/engines/openstack_kubespray.py
@@ -616,15 +616,22 @@ class OpenStack:
         self.meta["dns"] = [validate_ip(ip) for ip in
                             self.cluster.metadata.get("dns_nameservers", []).split(",")]
 
+        try:
+            self.meta["image"] = self.c.get_image(self.cluster.metadata["image_name"])
+            if self.meta["image"] is None:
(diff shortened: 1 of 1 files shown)


kqueen/engines/openstack_kubespray.py
```diff
@@ -586,15 +586,21 @@ def validate_ip(address):
...
                             self.cluster.metadata.get("dns_nameservers", []).split(",")]
 
+        try:
+            self.meta["image"] = self.c.get_image(self.cluster.metadata["image_name"])
+            if self.meta["image"] is None:
+                raise ValueError("Image '%s' is not found" % self.cluster.metadata["image_name"])
+            self.meta["flavor"] = self.c.get_flavor(self.cluster.metadata["flavor"])
+            if self.meta["flavor"] is None:
+                raise ValueError("Flavor '%s' is not found" % self.cluster.metadata["flavor"])
+            self.cluster.metadata["ssh_key_name"]
```
ekhomyakova: Please add a comment that these 2 lines are required to raise `KeyError` if the values are not presented in `self.cluster.metadata`.
Because at the first glance this code is a little bit confusing.

PR #338, opened 2018-07-09
Access ip workaround

redixin:
In some OpenStack deployments VM can't reach itself by floating ip. Some kubespray checks are not successful and deployment fails.

Solved by adding floating ip to loopback interface.

Also few minor fixes.

### DIFF
--- a/kqueen/engines/openstack_kubespray.py
+++ b/kqueen/engines/openstack_kubespray.py
@@ -393,8 +393,27 @@ class Kubespray:
         self._save_inventory(inventory, "hosts.json")
         self._create_group_vars()
         self._wait_for_ping()
+        self._add_fip_to_lo(resources)
         self._run_ansible()
-        return self._get_kubeconfig(resources["masters"][0]["ip"])
+        return self._get_kubeconfig(resources["masters"][0]["fip"])
+
+    def _add_fip_to_lo(self, resources):
+        """Add floating ip to loopback interface.
+
+        This is workaround for buggy deployments where node can't reach
+        itself by floating ip.
+        """
+        cmd_fmt = (
+            "sudo /bin/sh -c 'cat > /etc/rc.local <<EOF\n"
+            "/sbin/ip addr add %s/32 dev lo\n"
+            "EOF'"
+        )
+        for master in resources["masters"]:
+            ip = master["fip"]
+            host = "@".join((self.ssh_username, ip))
+            ssh_cmd = ("ssh", host) + self.ssh_common_args
+            subprocess.check_call(ssh_cmd + (cmd_fmt % ip, ))
+            subprocess.check_call(ssh_cmd + ("sudo /bin/sh /etc/rc.local", ))
 
     def scale(self, resources):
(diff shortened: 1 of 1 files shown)


kqueen/engines/openstack_kubespray.py
```diff
@@ -700,7 +724,14 @@ def _get_userdata(self):
             "package_update": True,
             "packages": ["python"],
             "ssh_authorized_keys": [self.extra_ssh_key],
+            "write_files": [
+                {
+                    "content": '{"bip": "10.13.0.1/16"}',
```
naumvd95: maybe define it as kwarg and get it from create-cluster page ? imo it will be much better for future deployments

kqueen/engines/openstack_kubespray.py
```diff
@@ -80,7 +80,6 @@ class OpenstackKubesprayEngine(BaseEngine):
                 "help_message": "Must be odd number",
                 "validators": {
                     "required": True,
-                    "min": 3,
                     "parity": "odd",
```
katyafervent: why min is not needed?

kqueen/engines/openstack_kubespray.py
```diff
@@ -534,7 +552,7 @@ def _run_ansible(self, inventory="hosts.json", playbook="cluster.yml"):
         )
         pipe.wait()
         if pipe.returncode:
-            raise RuntimeError("Non zero exit status from ansible (%s)" % pipe.returncode)
+            logger.warn("Non zero exit status from ansible (%s)" % pipe.returncode)
```
katyafervent: warn is deprecated, please use warning instead https://docs.python.org/3/library/logging.html#logging.Logger.warning

kqueen/engines/openstack_kubespray.py
naumvd95: and also need to specify this feature in docs.rst

kqueen/engines/openstack_kubespray.py
redixin: Because kubespray can deploy cluster with 1 master node

kqueen/engines/openstack_kubespray.py
redixin: See line 744. We also need to do the same at least with `kube_service_addresses` and `kube_pods_subnet`

naumvd95: yep, i know, but imo it should be different changes, we may define docker networks in this PR, coz it directly depends on main PR idea (bugfix) and provide separate PR with kube-networks. 2nd case - is docker networks defining and kube-networks defining  doesnot depends on each other and should be tested separately

kqueen/engines/openstack_kubespray.py
katyafervent: So please remove or from validation on server-side https://github.com/redixin/kqueen/blob/58ee6824412c298919ef5db5d893e88a8d7256ab/kqueen/engines/openstack_kubespray.py#L599

PR #340, opened 2018-07-11
Define docker bip as kwarg

naumvd95:


naumvd95:
need to check before merge

PR #341, opened 2018-07-12
Make k8s images configurable

katyafervent:


kqueen/engines/openstack_kubespray.py
```diff
@@ -610,7 +616,10 @@ def validate_ip(address):
             ipaddress.ip_address(address)
             return address
 
-        self.meta["master_count"] = self.cluster.metadata["master_count"]
+        mc = self.cluster.metadata["master_count"]
+        if mc % 2 == 0:
+            raise ValueError("Master node count must be an odd number at least 3 or greater")
```
naumvd95: afaik its redundant?

naumvd95:
do we need to forward it in compose-configs?

kqueen/engines/openstack_kubespray.py
katyafervent: redundant with what?

katyafervent:
this config is used in compose-demo file. I suggest to keep all Mirantis-related params in a one config

kqueen/engines/openstack_kubespray.py
```diff
@@ -441,8 +441,14 @@ def _create_group_vars(self):
         src = os.path.join(self.kubespray_path, "inventory/sample/group_vars")
         dst = self._get_cluster_path("group_vars")
         shutil.copytree(src, dst)
+
         with open(os.path.join(dst, "all.yml"), "a") as all_yaml:
-            all_yaml.write("\ncloud_provider: openstack\n")
+            data_to_add = {'cloud_provider': 'openstack'}
```
naumvd95: imo unclear naming , maybe kubespray_configuration?

katyafervent: kubespray_configuration is located at "all.yml" and we are open this file for append and in data_to_add dictionary contains everything that will be appended

naumvd95:
OK, LGFM, we can merge it , if it tested properly

kqueen/config/demo.py
```diff
@@ -33,3 +33,16 @@ class Config(BaseConfig):
     # Creds for Kqueen Read-only user
     LDAP_DN = 'cn=admin,dc=example,dc=org'
     LDAP_PASSWORD = 'heslo123'
+
+    # Images for Kubespray (available images that can be provided is listed in the link below)
```
ekhomyakova: *are listed

PR #342, opened 2018-07-13
Use separate flavors for slave and master nodes

katyafervent:


kqueen/engines/openstack_kubespray.py
```diff
@@ -64,10 +64,20 @@ class OpenstackKubesprayEngine(BaseEngine):
                     "required": True,
                 },
             },
-            "flavor": {
+            "master_flavor": {
                 "type": "text",
                 "order": 30,
-                "label": "Flavor",
+                "label": "Master flavor",
+                "default": config.get("KS_MASTER_FLAVOR"),
```
naumvd95: imo its better to set defaults directly here, for ex m1.medium for master, m1.large for slaves

kqueen/engines/openstack_kubespray.py
```diff
@@ -620,9 +630,12 @@ def validate_ip(address):
...
                 raise ValueError("Image '%s' is not found" % self.cluster.metadata["image_name"])
-            self.meta["flavor"] = self.c.get_flavor(self.cluster.metadata["flavor"])
-            if self.meta["flavor"] is None:
-                raise ValueError("Flavor '%s' is not found" % self.cluster.metadata["flavor"])
+            self.meta["master_flavor"] = self.c.get_flavor(self.cluster.metadata["master_flavor"])
+            if self.meta["master_flavor"] is None:
+                raise ValueError("Flavor for master '%s' is not found" % self.cluster.metadata["master_flavor"])
+            self.meta["slave_flavor"] = self.c.get_flavor(self.cluster.metadata["slave_flavor"])
+            if self.meta["slave_flavor"] is None:
+                raise ValueError("Flavor for slave '%s' is not found" % self.cluster.metadata["master_flavor"])
```
naumvd95: in what case it can be None? through api calls?

kqueen/engines/openstack_kubespray.py
```diff
@@ -620,9 +630,12 @@ def validate_ip(address):
             self.meta["image"] = self.c.get_image(self.cluster.metadata["image_name"])
             if self.meta["image"] is None:
                 raise ValueError("Image '%s' is not found" % self.cluster.metadata["image_name"])
-            self.meta["flavor"] = self.c.get_flavor(self.cluster.metadata["flavor"])
-            if self.meta["flavor"] is None:
-                raise ValueError("Flavor '%s' is not found" % self.cluster.metadata["flavor"])
+            self.meta["master_flavor"] = self.c.get_flavor(self.cluster.metadata["master_flavor"])
```
naumvd95: btw, why some checks are not in try-catch block?

kqueen/engines/openstack_kubespray.py
katyafervent: yes, it is api server) we need to validate data on server side too

kqueen/engines/openstack_kubespray.py
katyafervent: they are optional

kqueen/engines/openstack_kubespray.py
katyafervent: decided not to use values from config at all. may be later use profile for each provisioner with predefined settings

PR #343, opened 2018-07-13
Store kqueen data under /opt directory

katyafervent:
It will make files persistent, required for cluster management

### DIFF
--- a/kqueen/config/base.py
+++ b/kqueen/config/base.py
@@ -15,8 +15,8 @@ class BaseConfig:
     POOL_MAX_WORKERS = 64
 
     # Kubespray settings
-    KS_FILES_PATH = "/tmp/kqueen"
-    KS_KUBESPRAY_PATH = "/tmp/kubespray"
+    KS_FILES_PATH = "/opt/kqueen"
+    KS_KUBESPRAY_PATH = "./kubespray"
     KS_SSH_CMD = "/usr/bin/ssh"
     KS_SSH_KEYGEN_CMD = "/usr/bin/ssh-keygen"
     KS_ANSIBLE_CMD = "/usr/bin/ansible"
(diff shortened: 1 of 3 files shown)


naumvd95:
good idea and good job!
but i think we should change dir-names also  and add kqueen-ui-naming to dir-names, coz it really hard to find useful stack-config 
```
root@api-7b6759c9c4-kf7l8:/tmp/kqueen# ls -al
total 28
drwxr-xr-x 5 root root 4096 Jul 13 14:49 .
drwxrwxrwt 4 root root 4096 Jul 13 14:53 ..
drwxr-xr-x 3 root root 4096 Jul 13 11:14 0dafd4c5-42c2-4b6b-ad3a-72e4ccaf1783
drwxr-xr-x 3 root root 4096 Jul 13 13:40 8f16dca1-feea-4f00-b8a4-565a2d4cd13c
drwxr-xr-x 3 root root 4096 Jul 13 14:50 96dc75f6-f8a3-4ad9-aa20-b2d61321778e
-rw------- 1 root root 1675 Jul 12 10:17 ssh_key
-rw-r--r-- 1 root root  407 Jul 12 10:17 ssh_key.pub
```
3 clusters -> 3 different id's , but on UI side we can get only names
image

------------
so IMO we need to:
1) define ids in Cluster info on UI-side
or
2) change dir-names and add cluster-name as prefix/postfix/etc
@katyafervent  what do u think about it?

katyafervent:
Vlad, cluster name may contain symbols that are forbidden as directory names. regular user doesn't need to go to the kqueen directory. but you can find cluster id at the cluster detail link or send api call to get cluser info by specified id or get it from the etcd directly

naumvd95:
>>but you can find cluster id at the cluster detail link 
 it will be great!)

kqueen/config/base.py
```diff
@@ -15,8 +15,8 @@ class BaseConfig:
     POOL_MAX_WORKERS = 64
 
     # Kubespray settings
-    KS_FILES_PATH = "/tmp/kqueen"
-    KS_KUBESPRAY_PATH = "/tmp/kubespray"
+    KS_FILES_PATH = "/opt/kqueen"
+    KS_KUBESPRAY_PATH = "kubespray"
```
naumvd95: move in into /opt dir,  its better to define absolute path

PR #345, opened 2018-07-17
Save ansible logs while deploying cluster

katyafervent:
Logs for a particular cluister, deployed with OS-Kubespray provisioner
are now stored at the corresponding cluster directory under /opt/kqueen


kqueen/engines/openstack_kubespray.py
```diff
@@ -569,19 +568,18 @@ def _run_ansible(self, inventory="hosts.json", playbook="cluster.yml"):
...
-            stdin=subprocess.DEVNULL,
-            stdout=sys.stdout,
-            stderr=sys.stderr,
-            cwd=self.kubespray_path,
-            env=env,
-        )
-        pipe.wait()
-        if pipe.returncode:
-            logger.warning("Non zero exit status from ansible (%s)" % pipe.returncode)
+        with open(os.path.join(self._get_cluster_path(), "ansible_log.txt"), "a+") as ansible_log:
```
naumvd95: why we can't log it into common log?

katyafervent: it is ansible entity, we are logging only if something went wrong

PR #346, opened 2018-07-20
Improvements for OS-kubespray resize

katyafervent:


kqueen/engines/openstack_kubespray.py
```diff
@@ -334,10 +345,15 @@ def resize(self, node_count):
         node_count = int(node_count)
         master_count = len(self.cluster.metadata["resources"]["masters"])
         new_slave_count = node_count - master_count
-        if new_slave_count < 0:
-            return False, "Node count should be at least %s" % master_count
+        if new_slave_count <= 0:
+            return False, "Node count should be at least %s" % (master_count + 1)
```
naumvd95: maybe ``Node amount should not be less that Master-node count ``

kqueen/engines/openstack_kubespray.py
```diff
@@ -334,10 +345,15 @@ def resize(self, node_count):
...
         current_slave_count = len(self.cluster.metadata["resources"]["slaves"])
         delta = new_slave_count - current_slave_count
+
+        if delta == 0:
+            return False, "Cluster already has %s nodes" % node_count
+        self.cluster.state = config.CLUSTER_UPDATING_STATE
+        self.cluster.save()
```
naumvd95: raw 361, maybe just else?

naumvd95: elif redundant)

kqueen/engines/openstack_kubespray.py
```diff
@@ -592,8 +607,11 @@ def _run_ansible(self, inventory="hosts.json", playbook="cluster.yml"):
                 env=env,
             )
             pipe.wait()
-            if pipe.returncode:
-                logger.warning("Non zero exit status from ansible (%s)" % pipe.returncode)
+        if pipe.returncode:
+            with open(self.ansible_log, "r") as log_file:
+                result = log_file.read()
+            if result.find("fatal:"):
```
naumvd95: imo its bad case to find `fatal`. it looks like hardcode, can we trigger it on any other points? maybe exit code or smth similar?

kqueen/models.py
```diff
@@ -38,26 +38,25 @@ class Cluster(Model, metaclass=ModelMeta):
     owner = RelationField(required=True, remote_class_name='User')
 
     def update_state(self):
+        # Check for stale clusters
+        max_age = timedelta(seconds=config.get('PROVISIONER_TIMEOUT'))
```
naumvd95: what value do u prefer?

naumvd95: imo it can depends on many reasons, need to define to properly

kqueen/engines/openstack_kubespray.py
katyafervent: no, it is ansible expected behavior. we check error code first and then fatal, there are no any names from user input, so it is ok to look for fatal with :

kqueen/models.py
katyafervent: it is defined in the config file now, I didn't change that part

kqueen/engines/openstack_kubespray.py
katyafervent: Didn't get you, I didnt change 361 raw. Are you suggesting to to rename elif to else?

kqueen/engines/openstack_kubespray.py
katyafervent: May be yes, we need a better message here

naumvd95: agree, coz current message is unclear imo

kqueen/engines/openstack_kubespray.py
naumvd95: yep, i undrestand, its minor ) but if u have time )

kqueen/engines/openstack_kubespray.py
naumvd95: okok, agree

kqueen/models.py
naumvd95: ok, got it

PR #347, opened 2018-07-20
Get rid of pycrypto

katyafervent:


kqueen/storages/etcd.py
```diff
@@ -124,13 +123,9 @@ def _get_encryption_key(self):
...
 
         # Calculate hash password
-        return hashlib.sha256(key.encode('utf-8')).digest()[:self.bs]
-
-    def _pad(self, s):
-        return s + (self.bs - len(s) % self.bs) * chr(self.bs - len(s) % self.bs)
-
-    def _unpad(self, s):
-        return s[:-ord(s[len(s) - 1:])]
+        key = hashlib.md5(key.encode('utf-8')).hexdigest()
```
redixin: Why changed sha256 to md5?

kqueen/storages/etcd.py
```diff
@@ -110,11 +109,11 @@ def validate(self):
         """
         return True
 
-    def _get_encryption_key(self):
-        """Read encryption key and format it.
+    def _get_fernet_obj(self):
```
redixin: You may use AES instead of Fernet to be compatible with old database.

https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/

PR #350, opened 2018-07-27
Let user choose to use persistent volumes

katyafervent:
[OS-kubespray cluster] User is able to select persistent class usage
while creating cluster in OS provisioner

### DIFF
--- a/kqueen/engines/openstack_kubespray.py
+++ b/kqueen/engines/openstack_kubespray.py
@@ -131,6 +131,12 @@ class OpenstackKubesprayEngine(BaseEngine):
                 "label": "Availability zone",
                 "default": "nova",
             },
+            "persistent_volume": {
+                "type": "checkbox",
+                "order": 90,
+                "label": "Add Persistent Volumes Storage Class",
+                "help_message": "Default OpenStack Cinder Storage Class"
+            }
         },
         "provisioner": {
             "auth_url": {
@@ -238,7 +244,8 @@ class OpenstackKubesprayEngine(BaseEngine):
             self.cluster.metadata["node_count"] = node_count
             self.cluster.save()
             self.allow_addresses_for_calico(resources)
-            kubeconfig = self.ks.deploy(resources)
+
+            kubeconfig = self.ks.deploy(self.cluster.metadata)
             self.cluster.kubeconfig = kubeconfig
             self.cluster.state = config.CLUSTER_OK_STATE
             logger.info("Cluster provision completed")
@@ -423,10 +430,11 @@ class Kubespray:
                                 "-i", os.path.join(clusters_path, "ssh_key"))
         self._make_files_dir()
 
-    def deploy(self, resources):
+    def deploy(self, cluster_metadata):
+        resources = cluster_metadata['resources']
         inventory = self._generate_inventory(resources)
         self._save_inventory(inventory, "hosts.json")
-        self._create_group_vars()
+        self._create_group_vars(cluster_metadata)
         self._wait_for_ping()
         self._add_fip_to_lo(resources)
         self._run_ansible()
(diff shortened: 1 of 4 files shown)


naumvd95:
@katyafervent any news?

naumvd95:
@katyafervent what about that? we strongly need that case

kqueen/engines/openstack_kubespray.py
```diff
@@ -120,6 +120,12 @@ class OpenstackKubesprayEngine(BaseEngine):
                 "label": "Availability zone",
                 "default": "nova",
             },
+            "persistent_volume": {
+                "type": "checkbox",
+                "order": 90,
+                "label": "Add Persistent Volumes Storage Class",
```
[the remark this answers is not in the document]
katyafervent: it is set to false by default, if you need it, you can set checkbox to true

kqueen/engines/openstack_kubespray.py
```diff
@@ -477,18 +461,29 @@ def _save_inventory(self, inventory, filename):
...
-        with open(os.path.join(dst, "all.yml"), "a") as all_yaml:
-            data_to_add = {'cloud_provider': 'openstack'}
-            image_var_names = [var_name for var_name in dir(config) if var_name.endswith(('_IMAGE_REPO', '_IMAGE_TAG'))]
-            image_variables = {k.lower(): getattr(config, k) for k in image_var_names}
-            data_to_add.update(image_variables)
+        persistent_volumes = metadata.get("persistent_volume")
+        kubespray_vars = {"persistent_volumes_enabled": True} if persistent_volumes else {}
+
+        kubespray_vars["cloud_provider"] = "openstack"
+        kubespray_vars["openstack_blockstorage_version"] = "v2"
```
naumvd95: should be configurable

naumvd95: through config file at least

kqueen/engines/openstack_kubespray.py
```diff
@@ -477,18 +461,29 @@ def _save_inventory(self, inventory, filename):
...
+        kubespray_vars["calico_endpoint_to_host_action"] = "ACCEPT"
+        kubespray_vars["no_proxy"] = "127.0.0.1,localhost,docker-prod-local.docker.mirantis.net,172.16.48.254," \
+                                     "cloud-cz.bud.mirantis.net,172.17.45.80"
```
naumvd95: ditto, should be defined in mirantis.demo

kqueen/engines/openstack_kubespray.py
```diff
@@ -477,18 +461,29 @@ def _save_inventory(self, inventory, filename):
...
+
+        kubespray_vars["openstack_lbaas_subnet_id"] = metadata["resources"]["subnet_id"]
+        kubespray_vars["openstack_lbaas_floating_network_id"] = metadata["resources"]["ext_net_id"]
+        kubespray_vars["volume_cross_zone_attachment"] = True
```
naumvd95: should be configurable
 through config file at least

kqueen/engines/openstack_kubespray.py
```diff
@@ -708,6 +703,7 @@ def provision(self):
         router = self.c.create_router(name=self.stack_name,
                                       ext_gateway_net_id=self.meta['ext_net'].id)
         self.c.add_router_interface(router, subnet["id"])
+        resources["ext_net_id"] = self.meta['ext_net'].id
```
naumvd95: why single quotes?

kqueen/kubeapi.py
```diff
@@ -51,6 +51,8 @@ def get_api_client(self):
         # functions in https://github.com/kubernetes-client/python
         client_config = type.__call__(client.Configuration)
         kubeconfig = self.cluster.get_kubeconfig()
+        if kubeconfig is None:
```
naumvd95: maybe resolbve it with try catch block in self.cluster.get_kubeconfig method? imo it should be better

kqueen/engines/openstack_kubespray.py
naumvd95: oh, ok so forward that info here too
```## Running on top of openstack vms with cinder enabled may lead to unschedulable pods due to NoVolumeZoneConflict restriction in kube-scheduler.
## See https://github.com/kubernetes-incubator/kubespray/issues/2141
## Set this variable to true to get rid of this issue```

File: kqueen/kubeapi.py (hunk 6 above)
Comment (katyafervent) [reply kqueen/kubeapi.py]: get_kubeconfig just returns kubeconfig, it should not raise any error

Comment (naumvd95) [reply kqueen/kubeapi.py]: in what case it can be None? get_kubeconfig is really important for 70% kqueen actions, so why it can be None?

Comment (katyafervent) [reply kqueen/kubeapi.py]: If cluster has failed state

Comment (naumvd95) [reply kqueen/kubeapi.py]: ok, agree

File: kqueen/engines/openstack_kubespray.py [hunk 7]
```diff
@@ -477,18 +461,31 @@ def _save_inventory(self, inventory, filename):
...
         dst = self._get_cluster_path("group_vars")
         shutil.copytree(src, dst)
 
-        with open(os.path.join(dst, "all.yml"), "a") as all_yaml:
-            data_to_add = {'cloud_provider': 'openstack'}
-            image_var_names = [var_name for var_name in dir(config) if var_name.endswith(('_IMAGE_REPO', '_IMAGE_TAG'))]
-            image_variables = {k.lower(): getattr(config, k) for k in image_var_names}
-            data_to_add.update(image_variables)
+        persistent_volumes = metadata.get("persistent_volume")
+        kubespray_vars = {"persistent_volumes_enabled": True} if persistent_volumes else {}
```
naumvd95: ``kubespray_vars = {"persistent_volumes_enabled": True} if metadata.get("persistent_volume") else {}``
or i missed , that u need this var?

kqueen/engines/openstack_kubespray.py
```diff
@@ -597,8 +594,9 @@ def _run_ansible(self, inventory="hosts.json", playbook="cluster.yml"):
         args = [
             config.KS_ANSIBLE_PLAYBOOK_CMD, "-b", "-i",
             inventory, playbook,
-            "--extra-vars", "delete_nodes_confirmation=yes",
-            "--extra-vars", "docker_dns_servers_strict=no",
+            "-e", "delete_nodes_confirmation=yes",
```
naumvd95: why we need this? seems it can break deprovision actions

katyafervent: How? Here I've just add "ansible_python_interpreter=/usr/bin/python3" and use '-e' instead of '--extra-vars' just to make it more compact

kqueen/engines/openstack_kubespray.py
katyafervent: This far is responsible for storage class, other ones for the whole deployment (set in a separate commit)

PR #351, opened 2018-08-03
[OS-Kubespray] Define custom ansible configuration

naumvd95:


katyafervent:
Fixed in https://github.com/Mirantis/kqueen/pull/352

PR #353, opened 2018-08-10
[OS-Kubespray] Set up security group for openstack instances

katyafervent:


kqueen/engines/openstack_kubespray.py
```diff
@@ -20,6 +20,9 @@
 logger = logging.getLogger("kqueen_api")
 config = current_config()
 
+MASTER_SECURITY_GR = "kqueen_master"
+COMMON_SECURITY_GR = "kqueen_common"
```
naumvd95: what is it? why its hardcoded? if it permanent change, we should document it and provide possibility of change

kqueen/engines/openstack_kubespray.py
```diff
@@ -807,7 +813,40 @@ def _get_userdata(self):
         }
         return "#cloud-config\n" + yaml.dump(userdata)
 
-    def _boot_servers(self, *, name, servers_range, image, flavor, network,
+    def _set_up_security_groups(self):
+        master_sg = self.c.get_security_group(MASTER_SECURITY_GR)
```
naumvd95: imo all getters should be in one place in `init` method like any previous engines

kqueen/engines/openstack_kubespray.py
```diff
@@ -807,7 +813,40 @@ def _get_userdata(self):
...
+        if not master_sg:
+            master_sg = self.c.create_security_group(name=MASTER_SECURITY_GR,
```
naumvd95: imo we need a try/catch here

kqueen/engines/openstack_kubespray.py
```diff
@@ -807,7 +813,40 @@ def _get_userdata(self):
...
+                                              port_range_min="6443",
+                                              port_range_max="6443")
+            # Calico
+            self.c.create_security_group_rule(master_sg.id, protocol="tcp",
+                                              port_range_min="179",
+                                              port_range_max="179")
+
+        common_sg = self.c.get_security_group(COMMON_SECURITY_GR)
+        if not common_sg:
+            common_sg = self.c.create_security_group(name=COMMON_SECURITY_GR,
```
naumvd95: ditto

kqueen/engines/openstack_kubespray.py
naumvd95: imo we should provide an option on UI , smth like 'create  personal SC' otherwise it should use 'default'
its just one of useful ideas, need to talk about it with u

kqueen/engines/openstack_kubespray.py
```diff
@@ -807,7 +813,40 @@ def _get_userdata(self):
...
+
+            # Kubelet API
+            self.c.create_security_group_rule(common_sg.id, protocol="tcp",
+                                              port_range_min="10250",
+                                              port_range_max="10255")
+            # NodePort Services
+            self.c.create_security_group_rule(common_sg.id, protocol="tcp",
+                                              port_range_min="30000",
+                                              port_range_max="32767")
+        return master_sg, common_sg
```
naumvd95: there is no 'deprovision' step for SC

kqueen/engines/openstack_kubespray.py
katyafervent: without specific SG, openstack provisioner will not work

naumvd95: so need to document it as 'pre-step' or provide kqueen-OPS for that

kqueen/engines/openstack_kubespray.py
katyafervent: in init method data, collected from user is evaluated

kqueen/engines/openstack_kubespray.py
katyafervent: error catching is made in the above level (_run_provisioning function)

kqueen/engines/openstack_kubespray.py
katyafervent: you will not be able to delete sg if other cluster are using it. I think it will increase deletion time witch is so big now, but sg is not occupied space so fr me it is ok to leave it

naumvd95:
@katyafervent  what we should do with that PR? if we delay it, pls set additional labels

PR #356, opened 2018-08-31
Forbid deletion of current user for admins

ekhomyakova:


kqueen/blueprints/api/views.py
```diff
@@ -546,6 +546,20 @@ def get_policy_key(self):
         policy_key = super().get_policy_key()
         return '{}_{}'.format(policy_key, self.obj.role)
 
+    def dispatch_request(self, *args, **kwargs):
```
katyafervent: 1) I suggest to use super
2) 500 is server error response code, for forbidden actions I suggest to use 403: 

if str(kwargs['pk']) == current_identity.id:
            abort(403, "You can not delete yourself")
super().dispatch_request()

ekhomyakova: Done

PR #357, opened 2018-09-06
Improve os kubespray

katyafervent:


kqueen/engines/openstack_kubespray.py
```diff
@@ -711,6 +716,9 @@ def provision(self):
                                          image=self.meta['image'],
                                          flavor=self.meta['master_flavor'],
                                          network=network):
+            if master.status == 'ERROR':
+                raise RuntimeError('Could not spawn the instance. Check Openstack logs')
```
naumvd95: specify master.id or smth similar to point user into specific instance logs

kqueen/engines/openstack_kubespray.py
```diff
@@ -725,6 +733,8 @@ def provision(self):
                                         flavor=self.meta['slave_flavor'],
                                         network=network,
                                         add_random_suffix=True):
+            if slave.status == 'ERROR':
+                raise RuntimeError('Could not spawn the instance. Check Openstack logs')
```
naumvd95: specify slave.id or smth similar to point user into specific instance logs

PR #359, opened 2018-09-18
Fix getting status for OS-provisioner

katyafervent:


kqueen/engines/openstack_kubespray.py
```diff
@@ -822,7 +824,7 @@ def shrink(self, *, resources, remove_hostnames):
         return resources
 
     def _get_userdata(self):
-        docker_bip = {"bip": self.meta["docker_bip_network"]}
+        docker_bip = {"bip": self.cluster.metadata.get("docker_bip_network", "10.13.0.1/16")}
```
naumvd95: if u want to define "10.13.0.1/16" as sane default, please define it in kqueen config base.py

kqueen/engines/openstack_kubespray.py
```diff
@@ -756,6 +754,10 @@ def deprovision(self, volume_names):
                         self.c.remove_router_interface(router, port_id=i.id)
                         time.sleep(2)
                     self.c.delete_router(router.id)
+                    logger.debug("Router {0} has been deleted".format(self.stack_name))
+                if "resources" not in self.cluster.metadata:
+                    # Deployment failed before instance spawning
```
naumvd95: logging? exception?

katyafervent: exception was raised while provisioning, in that case we are handling deprovision of failed deployment

naumvd95: ok, agree

PR #361, opened 2018-09-27
Define floating ip's into k8s-config ssl context

naumvd95:
to avoid https://github.com/kubernetes-incubator/kubespray/issues/1430


kqueen/engines/openstack_kubespray.py
```diff
@@ -478,9 +478,20 @@ def _create_group_vars(self, metadata):
...
-        # Set this variable to true to get rid of this issue```
+        # Set this variable to true to get rid of this issue
         kubespray_vars["volume_cross_zone_attachment"] = True
 
+
+        # See https://github.com/kubernetes-incubator/kubespray/issues/1430
+        # Set all fips in this var to get rid of kubectl ssl-certs issue
+        ssl_fips = []
+        for master in metadata["resources"]["masters"]:
+            master_fip = master.get("fip", None)
```
katyafervent: we set 'fip' to master anyway, so it an be none) and this line is redundant

